PR #3, opened 2023-09-12
Add Kinesis source

oguzhanunlu:


snowplow-common-internal/kinesis/src/main/scala/com.snowplowanalytics.snowplow/sources/KinesisSource.scala
```diff
@@ -0,0 +1,252 @@
...
+      val hostname = InetAddress.getLocalHost().getCanonicalHostName()
+
+      val configsBuilder =
+        new ConfigsBuilder(
+          kinesisConfig.streamName,
+          kinesisConfig.appName,
+          kinesisClient,
+          dynamoDbClient,
+          cloudWatchClient,
+          s"$hostname:$uuid",
```
istreeter, line 177: Do we need the hostname here?  I checked what [fs2-aws does by default](https://github.com/laserdisc-io/fs2-aws/blob/v6.0.3/fs2-aws-kinesis/src/main/scala/fs2/aws/kinesis/Kinesis.scala#L146) does by default and it simply uses a uuid.

snowplow-common-internal/kinesis/src/main/scala/com.snowplowanalytics.snowplow/sources/KinesisSource.scala
```diff
@@ -0,0 +1,252 @@
...
+        LowLevelEvents(chunk.toList.map(getPayload), chunk.toList.map(toMetadata))
+      }
+  }
+
+  def getPayload(record: CommittableRecord): Array[Byte] = {
+    val data = record.record.data
+    val buffer = ArrayBuffer[Byte]()
+    while (data.hasRemaining)
+      buffer.append(data.get)
+    buffer.toArray
```
istreeter, line 154: Just pointing out that [rdb transformer has a slightly different implementation](https://github.com/snowplow/snowplow-rdb-loader/blob/5.7.3/modules/aws/src/main/scala/com/snowplowanalytics/snowplow/rdbloader/aws/Kinesis.scala#L96-L97)

Which is better?  I _guess_ it's the rdb-transformer implementation, because it copies in bulk instead of byte-by-byte.  The difference is important, because we'll do this for every single snowplow event.

**Idea** -  Both implementations are copying data from a ByteBuffer to an array.  This is expensive -- especially if we do it again and again for each event.  How about we set things up so that we don't need to copy the data!

We could change the definition of [TokenedEvents](https://github.com/snowplow-incubator/snowplow-lake-loader/blob/0.1.0/snowplow-common-internal/streams-core/src/main/scala/com.snowplowanalytics.snowplow/sources/TokenedEvents.scala#L22) so that `events` is a `List[ByteBuffer]`?  I think the downstream application can still work with that type, and it means we never need to copy bytes.

snowplow-common-internal/kinesis/src/main/scala/com.snowplowanalytics.snowplow/sources/KinesisSource.scala
```diff
@@ -0,0 +1,252 @@
...
+  private def lowLevel[F[_]: Parallel: Async](config: KinesisSourceConfig): LowLevelSource[F, List[KinesisMetadata[F]]] =
+    new LowLevelSource[F, List[KinesisMetadata[F]]] {
+      def checkpointer: KinesisCheckpointer[F] = kinesisCheckpointer[F]
+
+      def stream: Stream[F, Stream[F, LowLevelEvents[List[KinesisMetadata[F]]]]] =
+        Stream.emit(kinesisStream(config))
+    }
+
+  private def kinesisCheckpointer[F[_]: Parallel: Sync]: KinesisCheckpointer[F] = new KinesisCheckpointer[F] {
+    def combine(x: List[KinesisMetadata[F]], y: List[KinesisMetadata[F]]): List[KinesisMetadata[F]] = x ::: y
```
istreeter, line 77: Here goes my biggest criticism of the entire PR.....

Here, your checkpointer is accumulating a `List[KinesisMetadata[F]]`, which means you are storing a `KinesisMetadata[F]` in memory for every single event (until we checkpoint and clear the list).

For our applications which work in windows (lake loader, BRA, rdb transformer) there might be *millions* of events processed in between checkpoints.  So the list might grow to millions long.

It's not a huge problem -- we probably won't run out of memory -- but it's also completely unnecessary.  We only need to store in memory one `KinesisMetadata[F]` _per shard_.  Indeed, you later [drop and ignore the majority of the list](https://github.com/snowplow-incubator/snowplow-lake-loader/blob/e979ba15671e6ba3ca87e12280bb8701aeba2077/snowplow-common-internal/kinesis/src/main/scala/com.snowplowanalytics.snowplow/sources/KinesisSource.scala#L84).

How about, instead of

```scala
type KinesisCheckpointer[F[_]] = Checkpointer[F, List[KinesisMetadata[F]]]
```

...try to build it using

```scala
type KinesisCheckpointer[F[_]] = Checkpointer[F, Map[String, KinesisMetadata[F]]]
```

...where the map keys are the shard IDs.  This is approximately [what the RDB transformer does](https://github.com/snowplow/snowplow-rdb-loader/blob/5.7.3/modules/transformer-kinesis/src/main/scala/com/snowplowanalytics/snowplow/rdbloader/transformer/stream/kinesis/KinesisCheckpointer.scala#L23), although in other ways the transformer implementation is slightly different.

istreeter (commented):
Hey @oguzhanunlu this is very very good indeed!  I made a lot of comments, but only because this is a super-important module, so it's extremely important we get it right.  Don't be disheartened :)

I have not yet reviewed the integration tests.  @benjben are you interested in reviewing the integration tests?  Because you are more knowledgeable than me on how we like to use localstack and testcontainers

snowplow-common-internal/kinesis/src/main/scala/com.snowplowanalytics.snowplow/sources/KinesisSourceConfig.scala
```diff
@@ -0,0 +1,96 @@
...
+                              Latest.asRight
+                            case InitPositionRaw("AT_TIMESTAMP", Some(timestamp)) =>
+                              AtTimestamp(timestamp).asRight
+                            case other =>
+                              DecodingFailure(
+                                s"Initial position $other is not supported. Possible types are TRIM_HORIZON, LATEST and AT_TIMESTAMP (must provide timestamp field)",
+                                cur.history
+                              ).asLeft
+                          }
+        } yield initPosition
```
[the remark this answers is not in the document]
oguzhanunlu: yeah I took it from enrich, thanks both!

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/Utils.scala
```diff
@@ -0,0 +1,86 @@
...
+    _.evalMap { case TokenedEvents(events, token) =>
+      for {
+        _ <- ref.update(_ ::: events.map(bytes => new String(bytes, StandardCharsets.UTF_8)))
+      } yield token
+    }
+
+  def getLocalstackContainer(
+    exposedPort: Int,
+    streamName: String,
+    shardCount: Int
```
benjben, line 38: What about putting a default of `1` ?

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/Utils.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  def testProcessor(ref: Ref[IO, List[String]]): EventProcessor[IO] =
+    _.evalMap { case TokenedEvents(events, token) =>
+      for {
+        _ <- ref.update(_ ::: events.map(bytes => new String(bytes, StandardCharsets.UTF_8)))
+      } yield token
+    }
+
+  def getLocalstackContainer(
+    exposedPort: Int,
```
benjben, line 36: What about defining it in `LOCALSTACK_EXPOSED_PORT` in this file and using it as default ? If we have several tests, we still want only one localstack instance with the same exposed port.

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/Utils.scala
```diff
@@ -0,0 +1,86 @@
...
+    localstack.setWaitStrategy(Wait.forLogMessage(".*Ready.*", 1))
+    localstack
+  }
+
+  def getKinesisClient(endpoint: URI, region: Region): KinesisAsyncClient =
+    KinesisAsyncClient
+      .builder()
+      .endpointOverride(endpoint)
+      .region(region)
+      .build()
```
benjben, line 53: Can't this go wrong? Probably worth wrapping it in `IO`

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/Utils.scala
```diff
@@ -0,0 +1,86 @@
...
+    data: String
+  ): IO[PutRecordResponse] = {
+    val record = PutRecordRequest
+      .builder()
+      .streamName(streamName)
+      .data(SdkBytes.fromUtf8String(data))
+      .partitionKey(UUID.randomUUID().toString)
+      .build()
+
+    IO.blocking(client.putRecord(record).get())
```
benjben, line 67: Is there a particular reason to use an async client here given that we get the result directly and there are no callbacks?

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/KinesisSourceSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+      val region = Region.of(localstack.getRegion)
+      val config = getKinesisConfig(localstack.getEndpoint, localstack.getRegion, inputStreamName)
+
+      val examplePayload = "example-payload"
+
+      val prog = for {
+        refProcessed <- Ref[IO].of[List[String]](Nil)
+        _ <- putDataToKinesis(getKinesisClient(localstack.getEndpoint, region), inputStreamName, examplePayload)
+        sourceAndAck = KinesisSource.build[IO](config).stream(new EventProcessingConfig(NoWindowing), testProcessor(refProcessed))
+        fiber <- sourceAndAck.compile.drain.start
```
benjben, line 51:
```suggestion
        _ <- sourceAndAck.compile.drain
```
I'm curious about the benefit of using a `Fiber` here ?

benjben (changes requested):

snowplow-common-internal/kinesis/src/main/scala/com.snowplowanalytics.snowplow/sources/KinesisSource.scala
```diff
@@ -0,0 +1,252 @@
...
+  def getPayload(record: CommittableRecord): Array[Byte] = {
+    val data = record.record.data
+    val buffer = ArrayBuffer[Byte]()
+    while (data.hasRemaining)
+      buffer.append(data.get)
+    buffer.toArray
+  }
+
+  def getRuntimeRegion: String =
+    (new DefaultAwsRegionProviderChain).getRegion.id()
```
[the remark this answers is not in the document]
oguzhanunlu: yes `getRegion` can throw and I'll make it part of decoder

snowplow-common-internal/kinesis/src/main/scala/com.snowplowanalytics.snowplow/sources/KinesisSource.scala
oguzhanunlu: I liked the idea and making the required changes, thanks!

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/KinesisSourceSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+
+      val region = Region.of(localstack.getRegion)
+      val config = getKinesisConfig(localstack.getEndpoint, localstack.getRegion, inputStreamName)
+
+      val examplePayload = "example-payload"
+
+      val prog = for {
+        refProcessed <- Ref[IO].of[List[String]](Nil)
+        _ <- putDataToKinesis(getKinesisClient(localstack.getEndpoint, region), inputStreamName, examplePayload)
+        sourceAndAck = KinesisSource.build[IO](config).stream(new EventProcessingConfig(NoWindowing), testProcessor(refProcessed))
```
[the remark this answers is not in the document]
oguzhanunlu: hey @benjben , not sure if we miss that. `TimedWindows` is part of `EventProcessingConfig` which is downstream functionality and not a feature of `KinesisSource`. Why would `KinesisSourceSpec` test different processing approaches? Isn't it about whether `KinesisSource` can read from Kinesis or not?

This feels like duplicating `LowLevelSourceSpec` which already tests different window configs, what do you think?

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/KinesisSourceSpec.scala
oguzhanunlu: we need to terminate source stream so that we can end the test and check expectations

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/KinesisSourceSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+        fiber <- sourceAndAck.compile.drain.start
+        _ <- IO.sleep(2.minutes)
+        processed <- refProcessed.get
+        _ <- fiber.cancel
+      } yield processed must contain(examplePayload)
+
+      prog.unsafeRunSync()
```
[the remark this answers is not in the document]
oguzhanunlu: I tried it and timeout of `CatsEffect` caused a problem, perhaps there is a way out but I'll remove `CatsEffect` instead, since I need to control the timeout

snowplow-common-internal/kinesis/src/main/scala/com/snowplowanalytics/snowplow/sources/kinesis/KinesisSource.scala
```diff
@@ -0,0 +1,237 @@
...
+                )
+            }
+        }
+    def nack(c: Map[String, KinesisMetadata[F]]): F[Unit] = Applicative[F].unit
+  }
+
+  private def kinesisStream[F[_]: Async](config: KinesisSourceConfig): Stream[F, LowLevelEvents[Map[String, KinesisMetadata[F]]]] = {
+    val resources =
+      for {
+        region <- config.region.fold(KinesisSourceConfig.getRuntimeRegion.fold(Resource.raiseError[F, Region, Throwable], Resource.pure[F, Region]))(Resource.pure)
```
istreeter, line 116: This is.... ok.

It will work, and that's what I care about most.  So I don't want to complain too loudly.

But I will give some opinions:

First, I find it very hard to read anything that uses `option.fold(default)(f)` if `default` and `f` are themselves very long to read.  My eyes just can't scan it easily.  And code legibility is important, because your primary audience is your colleagues (the compiler is only the secondary audience).

IMO if something is hard to read, then it's worth a refactor out of kindness to your reviewer 😉

Next.....

This will cause an exception if the user forgot to configure `region` in their config file.  But that's just a config mistake, so it should be caught at config-parsing time.  So the `ConfigParser` module can give a kind message to the user explaining what went wrong and where.

I think that's a nicer experience for the user instead of getting an exception when the app tries to connect to the stream.

In other words... I would make `region` non-optional in the `KinesisSourceConfig` class.  And `getRuntimeRegion` can be part of the decoder.

snowplow-common-internal/kinesis/src/main/scala/com/snowplowanalytics/snowplow/sources/kinesis/KinesisSourceConfig.scala
```diff
@@ -0,0 +1,65 @@
...
+  sealed trait InitPosition
+
+  object InitPosition {
+    case object Latest extends InitPosition
+
+    case object TrimHorizon extends InitPosition
+
+    case class AtTimestamp(timestamp: Instant) extends InitPosition
+
+    implicit val decoder: Decoder[InitPosition] = deriveDecoder[InitPosition]
```
istreeter, line 51: Have you tried decoding anything yet to see if your decoder works?

(I think it should be `deriveConfiguredDecoder` from the `io.circe.generic.extras.semiauto` package.  But testing it out will tell you).

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/KinesisSourceSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+      val inputStreamName = "read-input-test-stream"
+      val shardCount = 1
+      val localstack = getLocalstackContainer(4566, inputStreamName, shardCount)
+      localstack.start()
+
+      val logger = LoggerFactory.getLogger("kinesis-source-test")
+      val logs = new Slf4jLogConsumer(logger)
+      localstack.followOutput(logs)
+
+      val region = Region.of(localstack.getRegion)
```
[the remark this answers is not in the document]
oguzhanunlu: I love aws sdk

```scala
scala> import software.amazon.awssdk.regions.Region
import software.amazon.awssdk.regions.Region

scala> Region.of("potato")
val res0: software.amazon.awssdk.regions.Region = potato

scala> res0.id
val res1: String = potato
```

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Processing.scala
```diff
@@ -151,17 +152,18 @@ object Processing {
...
       list
-        .traverse { bytes =>
+        .traverse { byteBuffer =>
           Applicative[F].pure {
-            val stringified = new String(bytes, StandardCharsets.UTF_8)
+            val stringified = StandardCharsets.UTF_8.decode(byteBuffer).toString
             Event
               .parse(stringified)
-              .map(event => Parsed(event, bytes.size, TabledEntity.forEvent(event)))
+              .map(event => Parsed(event, byteBuffer.array().length, TabledEntity.forEvent(event)))
```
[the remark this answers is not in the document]
oguzhanunlu: `array()` implementation doesn't seem to copy contents to a new array, instead it returns the byte array backing the byte buffer.

In this case, `array().length` and `limit` are the same thing, I'll switch to `limit` for safety

snowplow-common-internal/pubsub/src/main/scala/com.snowplowanalytics.snowplow/sources/PubsubSource.scala
```diff
@@ -94,7 +96,7 @@ object PubsubSource {
           LowLevelEvents(events, acks)
         }
         .evalTap { case LowLevelEvents(events, _) =>
-          val numBytes = events.map(_.size).sum
+          val numBytes = events.map(_.array().length).sum
```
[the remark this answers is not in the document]
oguzhanunlu: `array()` doesn't convert but return the backing array, both are same, I'll use `limit`

snowplow-common-internal/kinesis/src/main/scala/com/snowplowanalytics/snowplow/sources/kinesis/KinesisSource.scala
oguzhanunlu: appreciate this feedback, thanks

snowplow-common-internal/kinesis/src/main/scala/com.snowplowanalytics.snowplow/sources/KinesisSourceConfig.scala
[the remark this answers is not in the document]
oguzhanunlu: I got rid of intermediate representation of init position but I still needed custom decoder

snowplow-common-internal/kinesis/src/main/scala/com/snowplowanalytics/snowplow/sources/kinesis/KinesisSourceConfig.scala
oguzhanunlu: it wasn't working as expected and I needed custom decoder as seen in my last commits

snowplow-common-internal/kinesis/src/main/scala/com/snowplowanalytics/snowplow/sources/kinesis/KinesisSourceConfig.scala
```diff
@@ -0,0 +1,100 @@
...
+  retrievalMode: KinesisSourceConfig.Retrieval,
+  bufferSize: PosInt,
+  customEndpoint: Option[URI],
+  dynamodbCustomEndpoint: Option[URI],
+  cloudwatchCustomEndpoint: Option[URI]
+)
+
+object KinesisSourceConfig {
+
+  implicit val config: Configuration = Configuration.default.withDefaults
```
istreeter, line 36: I need to think about this... whether it's better to define the `Configuration` here, or whether it's better to pass a `Configuration` implicitly when creating the decoder.

I'll come back to this.

snowplow-common-internal/pubsub/src/main/scala/com.snowplowanalytics.snowplow/sources/PubsubSource.scala
```diff
@@ -94,7 +96,7 @@ object PubsubSource {
           LowLevelEvents(events, acks)
         }
         .evalTap { case LowLevelEvents(events, _) =>
-          val numBytes = events.map(_.size).sum
+          val numBytes = events.map(_.limit()).sum
```
istreeter, line 99: Just a warning, I fixed a bug on this line in the main branch: https://github.com/snowplow-incubator/snowplow-lake-loader/commit/95fab3c39507f69c23ff551b07564fded6946632

So we'll need to reconcile these two changes before we publish this as a libray.

One option is to rebase everything now onto `main`.  So we reconcile it before we do the copy&paste into the other repo.  WDYT?

istreeter (commented):
A lot of the methods in `KinesisSource` and `KinesisSourceConfig` can be private.  By making them private, we reduce the API surface of the library you are creating.  And then we have more flexibility to adapt the library in future without making breaking changes.

The only public methods should be any offically supported method that we want applications to use.

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/KinesisSourceSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+
+import software.amazon.awssdk.regions.Region
+
+import scala.annotation.nowarn
+import scala.concurrent.duration.DurationInt
+
+import com.snowplowanalytics.snowplow.sources.EventProcessingConfig.NoWindowing
+import com.snowplowanalytics.snowplow.sources.Utils._
+
+@nowarn("msg=unused value of type org.specs2.specification.core.Fragment")
```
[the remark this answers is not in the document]
istreeter: Bump this question from last week ^^

project/Dependencies.scala
```diff
@@ -124,6 +138,22 @@ object Dependencies {
...
+  val kinesisDependencies = Seq(
+    fs2AwsKinesis,
+    arnsSdk2,
+    kinesisSdk2,
+    dynamoDbSdk2,
+    cloudwatchSdk2,
+    circeConfig,
+    circeGeneric,
+    circeGenericExtra,
+    "io.circe"         %% "circe-parser" % V.circe,
```
istreeter, line 150: Is this line left here by accident?

project/Dependencies.scala
```diff
@@ -124,6 +138,22 @@ object Dependencies {
     slf4j % Provided
   )
 
+  val kinesisDependencies = Seq(
+    fs2AwsKinesis,
+    arnsSdk2,
```
istreeter, line 143: I removed this line and it still compiled.  Is it needed?

istreeter: Ignore!  Sorry my mistake I got this one wrong.

project/Dependencies.scala
oguzhanunlu: yes :) thanks for the heads up!

snowplow-common-internal/kinesis/src/main/scala/com/snowplowanalytics/snowplow/sources/kinesis/KinesisSourceConfig.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  implicit val posIntDecoder: Decoder[PosInt] =
+    Decoder.decodeInt.emap { int =>
+      if (int > 0) PosInt.from(int)
+      else s"Positive integer expected but [$int] is provided".asLeft[PosInt]
```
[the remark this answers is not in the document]
oguzhanunlu: definitely 👍

snowplow-common-internal/pubsub/src/main/scala/com.snowplowanalytics.snowplow/sources/PubsubSource.scala
oguzhanunlu: I agree, after code review is completed, I'll rebase this branch onto `main`

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/KinesisSourceSpec.scala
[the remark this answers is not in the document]
oguzhanunlu: thanks for the heads up, I just sent a new commit to address this

.scalafmt.conf
```diff
@@ -1,8 +1,31 @@
...
+    owners = [{
+      regex = "Term.ApplyInfix"
+    }]
+  }, {
+    code = "%%"
+    owners = [{
+      regex = "Term.ApplyInfix"
+    }]
+  }
+]
```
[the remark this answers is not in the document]
oguzhanunlu: this is just personal opinion and I don't think we have a team consensus regarding formatting afaik

yeah I should add sbt check 👍

snowplow-common-internal/kinesis/src/main/scala/com/snowplowanalytics/snowplow/sources/kinesis/KinesisSource.scala
```diff
@@ -0,0 +1,236 @@
...
+    LowLevelSource.toSourceAndAck(lowLevel(config))
+
+  private type KinesisCheckpointer[F[_]] = Checkpointer[F, Map[String, KinesisMetadata[F]]]
+
+  private implicit class RichCommitableRecord(val cr: CommittableRecord) extends AnyVal {
+    def toMetadata[F[_]: Sync]: KinesisMetadata[F] =
+      KinesisMetadata(cr.shardId, cr.sequenceNumber, cr.isLastInShard, cr.lastRecordSemaphore, cr.checkpoint)
+  }
+
+  final case class KinesisMetadata[F[_]](
```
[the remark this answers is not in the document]
oguzhanunlu: I wasn't sure about it at first but yeah, it should be private

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/kinesis/Resources.scala
```diff
@@ -0,0 +1,58 @@
...
+import org.testcontainers.utility.DockerImageName
+import software.amazon.awssdk.regions.Region
+import software.amazon.awssdk.services.kinesis.KinesisAsyncClient
+
+import java.net.URI
+import java.nio.charset.StandardCharsets
+
+object Resources {
+
+  def prepareLocalstackContainer(region: Region, kinesisInitializeStreams: String): LocalStackContainer = {
```
benjben, line 25:
```suggestion
  def resource(region: Region, kinesisInitializeStreams: String): Resource[LocalStackContainer] = {
```
IMHO this should create the localstack resource directly, so call the `start` and the `stop`.

I also think that this file should only contain localstack stuff, so it can probably be renamed `Localstack` and this function `resource`

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/kinesis/Resources.scala
```diff
@@ -0,0 +1,58 @@
...
+    val localstack = new LocalStackContainer(DockerImageName.parse("localstack/localstack:2.2.0"))
+    localstack.addEnv("AWS_DEFAULT_REGION", region.id)
+    localstack.addEnv("KINESIS_INITIALIZE_STREAMS", kinesisInitializeStreams)
+    localstack.addExposedPort(4566)
+    localstack.setWaitStrategy(Wait.forLogMessage(".*Ready.*", 1))
+    localstack
+  }
+
+  def startLocalstack(localstack: LocalStackContainer): LocalStackContainer = {
```
benjben, line 34:
```suggestion
  private def startLocalstack(localstack: LocalStackContainer): LocalStackContainer = {
```

snowplow-common-internal/kinesis/src/it/scala/com/snowplowanalytics/snowplow/sources/kinesis/Resources.scala
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+  def getKinesisClient(endpoint: URI, region: Region): IO[KinesisAsyncClient] =
+    IO(
+      KinesisAsyncClient
+        .builder()
+        .endpointOverride(endpoint)
+        .region(region)
+        .build()
+    )
```
benjben, line 56: Seems that these functions rather belong to `Utils`

benjben (approved):
I only reviewed the integration tests

snowplow-common-internal/kinesis/src/main/scala/com/snowplowanalytics/snowplow/sources/kinesis/KinesisSourceConfig.scala
```diff
@@ -0,0 +1,66 @@
...
+import io.circe._
+import io.circe.generic.extras.semiauto.deriveConfiguredDecoder
+import io.circe.generic.extras.Configuration
+import software.amazon.awssdk.regions.Region
+import software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain
+
+import java.net.URI
+import java.time.Instant
+
+case class KinesisSourceConfig(
```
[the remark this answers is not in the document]
oguzhanunlu: let's do it in the new lib

PR #6, opened 2023-09-12
Release/0.1.1

istreeter:


CHANGELOG
```diff
@@ -1,3 +1,8 @@
+Version 0.1.1 (2023-09-12)
```
oguzhanunlu, line 1: nit: date

oguzhanunlu (approved):
Thanks @istreeter ! It is not a blocker but I think the new unit test should be checking if recovery column is observed in output columns, if I understood the issue correctly

PR #15, opened 2023-09-19
Bug fixes and tidy up 0.1.2

istreeter:


istreeter:
@stanch this PR includes some tidy-up to the example config files.  To make them appropriate for import into the docs site.

Are you interested in reviewing these example config changes?  You don't need to review the other code changes, although you're welcome to if you want.

config/config.azure.reference.hocon
```diff
@@ -1,31 +1,31 @@
...
   "output": {
 
     "good": {
-      # -- uri of the bucket where the data lake will be written
-      # -- - For Azure blob storage, the uri should start with `abfs://`
-      # -- - For local development, this can be a path to a local file e.g. `file:///path/to/events`
-      "location": "abfs://container1@streeterdev1.dfs.core.windows.net/events"
+      # -- URI of the bucket where the data lake will be written (required)
+      # -- For Azure blob storage, the uri should start with `abfs://`
+      "location": "abfs://snowplow@example.dfs.core.windows.net/events"
```
stanch, line 24: What about abfss://? Should we not recommend the more secure option?

istreeter: Have you tested that it works already?  I wasn't aware of the difference.  But if you say it works with `abfss` then I'll definitely change it.

project/Dependencies.scala
```diff
@@ -97,16 +98,13 @@ object Dependencies {
   val thrift            = "org.apache.thrift"          % "libthrift"                          % V.thrift
   val jackson           = "com.fasterxml.jackson.core" % "jackson-databind"                   % V.jackson
 
-  // snowplow: Note jackson-databind 2.14.x is incompatible with Spark
   val badrows           = "com.snowplowanalytics"      %% "snowplow-badrows"                      % V.badrows
   val tracker           = "com.snowplowanalytics"      %% "snowplow-scala-tracker-core"           % V.tracker
   val trackerEmit       = "com.snowplowanalytics"      %% "snowplow-scala-tracker-emitter-http4s" % V.tracker
-  val schemaDdl         = ("com.snowplowanalytics"      %% "schema-ddl"               % V.schemaDdl)
-    .exclude("com.fasterxml.jackson.core", "jackson-databind")
```
spenes, line 105: Don't we need these exclusions anymore ?

spenes (approved):

project/Dependencies.scala
istreeter: No, we don't need them.  In a recent previous commit I bumped spark to a newer version, and the newer version is compatible with jackson 2.14.

CHANGELOG
```diff
@@ -1,3 +1,11 @@
+Version 0.1.3 (2023-09-21)
```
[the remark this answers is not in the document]
istreeter: Thank you! Yes you are correct.

PR #16, opened 2023-09-21
Use circe's auto decoders for sealed traits

istreeter:


istreeter:
@oguzhanunlu this is what I was thinking of when I said to use the auto-derived decoders instead of writing your own decoder.

I like this one because it removes a lot of the complexity compared to your decoders, which I found difficult to read.

I also added a Spec to demo how it works.

istreeter:
cc @benjben too as you've been involved in this discussion.

oguzhanunlu:
this is much more easier to read & understand, thanks @istreeter !

PR #19, opened 2023-10-14
Import common loaders lib from maven (close #18)

### LINKED ISSUE #18: Import common loaders lib from maven
This repo was hosting local copies of our common loaders classes.  But they have now moved to [the common-streams github repo](https://github.com/snowplow-incubator/common-streams) and are published to maven.

istreeter:


modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/Run.scala
```diff
@@ -48,18 +48,23 @@ object Run {
...
-        LogUtils.prettyLogException(e).as(ExitCode.Error)
-    }
+    eitherT
+      .leftSemiflatMap { s: String =>
+        Logger[F].error(s).as(ExitCode.Error)
+      }
+      .merge
+      .handleErrorWith { e =>
+        Logger[F].error(e)("Exiting") >>
+          LogUtils.prettyLogException(e).as(ExitCode.Error)
```
benjben, line 61: Are we not logging `e` twice ?

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Processing.scala
```diff
@@ -151,26 +152,27 @@ object Processing {
...
-                val payload = BadRowRawPayload(stringified)
-                BadRow.LoaderParsingError(processor, failure, payload)
-              }
-          }
+          for {
+            // order of these byte buffer operations is important
+            numBytes <- Sync[F].delay(bytes.limit() - bytes.position())
+            stringified <- Sync[F].delay(StandardCharsets.UTF_8.decode(bytes).toString)
+          } yield Event
+            .parse(stringified)
```
benjben, line 164: :art: format is weird

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Processing.scala
```diff
@@ -151,26 +152,27 @@ object Processing {
...
+            .map(event => Parsed(event, numBytes, TabledEntity.forEvent(event)))
+            .leftMap { failure =>
+              val payload = BadRowRawPayload(stringified)
+              BadRow.LoaderParsingError(processor, failure, payload)
```
benjben, line 168: Bad rows are never written to a stream and we don't need to check their size right?

project/Dependencies.scala
```diff
@@ -11,21 +11,13 @@ object Dependencies {
...
-    val cats             = "2.9.0"
     val catsEffect       = "3.5.0"
     val catsRetry        = "3.1.0"
-    val fs2              = "3.7.0"
-    val log4cats         = "2.6.0"
-    val http4s           = "0.23.15"
     val decline          = "2.4.1"
     val circe            = "0.14.1"
-    val circeConfig      = "0.10.0"
+    val http4s           = "0.23.15"
```
benjben, line 18: There is now `0.23.23`

benjben (approved):
:+1:

project/Dependencies.scala
istreeter: Thanks, will bump everything to latest versions before 0.2.0 release.

PR #24, opened 2023-10-26
Add latency statsd metrics (close #23)

### LINKED ISSUE #23: Add latency statsd metric
We want to consolidate all snowplow apps to emit a latency metric with a common definition:

```
latency =
  { time the app started processing the event } - { time the event was written to the input stream }
```

The common latency metric is basically a measure of whether the app is keeping up with the source of events or whether it is lagging.  The metric name will be `latency_millis`.

Before now, the lake loader had a different metric called `processing_latency_seconds` which was a measure of how long the app spent processing events internally.  This is still a helpful metric for different reasons, so we will keep it.

I plan to rename `processing_latency_seconds` to `processing_latency_millis` so that the two latency metrics have consistent units.  This is a breaking change, but we're only on version 0.1.0 so I that's OK.

istreeter:


modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/Metrics.scala
```diff
@@ -34,18 +35,20 @@ object Metrics {
...
         KVMetric.CountBad(bad),
-        KVMetric.CountCommitted(committed)
+        KVMetric.CountCommitted(committed),
+        KVMetric.Latency(latency)
       ) ++ processingLatency.map(KVMetric.ProcessingLatency(_))
   }
 
   private object State {
-    def empty: State = State(0, 0, 0, None)
+    def empty: State = State(0, 0, 0, Duration.Zero, None)
```
benjben, line 51: Why isn't the `latency` optional like `processing_latency` ?

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Processing.scala
```diff
@@ -129,6 +130,20 @@ object Processing {
       }
     }.drain
 
+  private def setLatency[F[_]: Sync](metrics: Metrics[F]): Pipe[F, TokenedEvents, TokenedEvents] =
+    _.evalTap {
```
benjben, line 134: Do we use a `Pipe` and `evalTap` instead of `observe` because the docs say that this is faster or is there another reason?

modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/MockEnvironment.scala
```diff
@@ -109,4 +120,23 @@ object MockEnvironment {
...
+      ref.update(_ :+ AddedCommittedCountMetric(count))
+
+    def setLatency(latency: FiniteDuration): IO[Unit] =
+      ref.update(_ :+ SetLatencyMetric(latency))
+
+    def setProcessingLatency(latency: FiniteDuration): IO[Unit] =
+      ref.update(_ :+ SetProcessingLatencyMetric(latency))
+
+    def report: Stream[IO, Nothing] = Stream.never[IO]
+  }
```
benjben, line 141: :ok_hand:

modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/processing/ProcessingSpec.scala
```diff
@@ -29,68 +32,106 @@ class ProcessingSpec extends Specification with CatsEffect {
...
+        Action.AddedCommittedCountMetric(4),
+        Action.SetProcessingLatencyMetric(MockEnvironment.WindowDuration),
         Action.Checkpointed(inputs.map(_.ack)),
         Action.RemovedDataFramesFromDisk(List("view-0"))
       )
     )
 
-  def e2 =
-    for {
+    TestControl.executeEmbed(io)
```
benjben, line 58: First time that I see the test kit in action, great!

benjben (approved):
Awesome!

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/Metrics.scala
istreeter: I'll explain but the explanation is not great!!

I implemented this in the snowflake loader first, before I did it for the lake loader.  And there I decided to make it non-optional because... I _think_ it makes sense to say latency is zero if there are no input events.  If there are no events to process, then it is not lagging, so latency is zero.  I decided it was helpful to emit the zero metric to show the app is healthy  (you could argue this differently though).

Then I did it for the lake loader, and I decided to make the implementation identical to the snowflake loader, so non-optional.

But... the lake loader already had a latency metric: the processing latency.  So why had I made that one optional?

The lake loader is unusual because it sinks events only once per 5 minutes.  So the processing latency (time from receiving to sinking) is like this:

```
undefined
undefined
undefined
undefined
300 (ish)
undefined
undefined
undefined
undefined
300 (ish)
```

For processing latency it did not feel right to say latency = 0 in the intervening minutes.  Because it's not really zero if it's still processing something.

So we've ended up with an inconsistency between the two latency metrics!  One is optional and one is not.  For now, I am OK with this.  And I am OK to revisit the decision once we actually start using these metrics for alarms, reporting and debugging.

PR #26, opened 2023-10-26
Set user-agent headers when using GCP SDKs

### LINKED ISSUE #25: Set user-agent headers when using GCP SDKs

istreeter:


modules/gcp/src/main/resources/application.conf
```diff
@@ -12,4 +12,10 @@
       "requestByteThreshold": 1000000
     }
   }
+
+  "spark": {
+    "conf": {
+      "fs.gs.storage.http.headers.user-agent": "snowplow-lake-loader/0.2.0 (GPN:snowplow;oss)"
```
jbeemster, line 18: It being setup like this ... is it possible for a user to remove this GPN?  Should we bake it deeper into the code itself over being able to configure all of it?

istreeter: > is it possible for a user to remove this GPN?

Yes!  For example, when running this on the BDP platform we ourselves will change it to something different.

> Should we bake it deeper into the code itself over being able to configure all of it?

This way is consistent with how RDB loader is configured for Snowflake and Databricks user agent headers.

An alternative is we force it to be either the OSS GPN or the BDP GPN.  We didn't do it that way in RDB loader because of mixing BDP-specific logic into open source software.

jbeemster: Fair enough - lets leave it flexible then and stay consistent with RDB Loader!

PR #33, opened 2023-11-26
Candidates 0.2.0

istreeter:


modules/core/src/main/resources/reference.conf
```diff
@@ -1,4 +1,11 @@
 {
+
+  "gcpUserAgent": {
+    "productName": "Snowplow OSS"
+    "productName": ${?GCP_USER_AGENT_PRODUCT_NAME}
```
spenes, line 5: I don't this syntax exactly but does it mean `productName` will be set to GCP_USER_AGENT_PRODUCT_NAME env variable if it is set ?

modules/core/src/main/resources/reference.conf
```diff
@@ -1,4 +1,11 @@
...
+    "productVersion": "lake-loader"
```
spenes, line 6: Seems like `productVersion` isn't used anywhere ?

spenes (approved):
Looks good! Left two small comments.

modules/core/src/main/resources/reference.conf
istreeter: In the gcp `application.conf` file [here](https://github.com/snowplow-incubator/snowplow-lake-loader/blob/927b799f5c2713dc35b446ff2fb047088f16503c/modules/gcp/src/main/resources/application.conf#L4) I copy this `gcpUserAgent` block into the pubsub source/sink configurations.  Then the common-streams implementations expects to see a `productVersion` field and it is used [here](https://github.com/snowplow-incubator/common-streams/blob/bc6efa7494d4699c80c63e34782c75b563239286/modules/pubsub/src/main/scala/com/snowplowanalytics/snowplow/pubsub/GcpUserAgent.scala#L22).

Sorry I know that is really not obvious.

modules/core/src/main/resources/reference.conf
istreeter: Yes.  I did it this way because otherwise we would need to change `productName` in three different places in the hocon file:

- `spark.gcpUserAgent`
- `input.gcpUserAgent`
- `output.bad.gcpUserAgent`

But by doing it this way we just need to set one environment variable.

In retrospect, I cannot decide if I like this or not.  It reduces the number of places to set the configuration, but adds some obscurity around how the configuration flows through.  What do you think?

spenes: I think it is okay to leave it like this.

PR #36, opened 2023-12-06
Custom AWS credentials provider with support for external ID (close #35)

### LINKED ISSUE #35: Custom AWS credentials provider with support for external ID
For the AWS Lake Loader, we want to assume a role when writing to S3.  The hadoop S3a filesystem allows [many methods](https://hadoop.apache.org/docs/stable/hadoop-aws/tools/hadoop-aws/index.html#Authenticating_with_S3) for authenticating, of which the `org.apache.hadoop.fs.s3a.auth.AssumedRoleCredentialProvider` [comes closest to being what we need](https://hadoop.apache.org/docs/stable/hadoop-aws/tools/hadoop-aws/assumed_roles.html).  Unfortunately, though, it does not allow us to configure an [external id](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html) when assuming the role.

For best security practice, we want S3 bucket owners to protect their data by requiring the loader to specify a unique external id when assuming the cross-account role.  So we need to implement our own credentials provider.  It can be a very lightweight wrapper around a delegate credentials provider.

istreeter:


spenes (approved):
LGTM!

PR #38, opened 2023-12-14
Integration tests for Hudi and Iceberg output formats

istreeter:
Before this PR, we had an integration test for Delta, but nothing for Iceberg and Hudi.

This helps with managing the project dependencies.  It is all too easy to exclude a runtime dependency and then find that loading has broken for one of the formats.

### DIFF
--- a/modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/TestConfig.scala
+++ b/modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/TestConfig.scala
@@ -11,23 +11,51 @@ import com.typesafe.config.ConfigFactory
 import cats.implicits._
 import io.circe.config.syntax._
 
+import fs2.io.file.Path
+
 object TestConfig {
 
+  sealed trait Target
+  case object Delta extends Target
+  case object Hudi extends Target
+  case object Iceberg extends Target
+
   /** Provides an app Config using defaults provided by our standard reference.conf */
-  def defaults(overrides: String = fallbacks): AnyConfig =
+  def defaults(target: Target, tmpDir: Path): AnyConfig =
     ConfigFactory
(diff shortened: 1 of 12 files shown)


modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/TestConfig.scala
```diff
@@ -14,20 +14,20 @@ import io.circe.config.syntax._
 object TestConfig {
 
   /** Provides an app Config using defaults provided by our standard reference.conf */
-  def defaults: AnyConfig =
+  def defaults(overrides: String = fallbacks): AnyConfig =
```
pondzix, line 17: Do we need fallback? It seems we always provide some overrides?

modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/TestSparkEnvironment.scala
```diff
@@ -28,9 +28,15 @@ case class TestSparkEnvironment(
...
 object TestSparkEnvironment {
 
-  def build(windows: List[List[TokenedEvents]]): Resource[IO, TestSparkEnvironment] = for {
+  sealed trait Target
+  case object Delta extends Target
+  case object Hudi extends Target
+  case object Iceberg extends Target
+
+  def build(target: Target, windows: List[List[TokenedEvents]]): Resource[IO, TestSparkEnvironment] = for {
     tmpDir <- Resource.eval(IO.blocking(Files.createTempDirectory("lake-loader")))
```
pondzix, line 37: Not related to this change, but maybe we could use here: `fs2.io.file.Files.tempDirectory`? Because at the moment this temp directory is not cleaned when test finishes, right?

PR #39, opened 2023-12-15
Seperate docker builds for non-prod lake formats

istreeter:


modules/aws/src/main/scala/com.snowplowanalytics.snowplow.lakes/AssumedRoleCredentialsProvider.scala
```diff
@@ -37,23 +49,35 @@ class AssumedRoleCredentialsProvider(delegate: StsAssumeRoleCredentialsProvider)
...
+
+  private def getDelegate(conf: Configuration): AwsCredentialsProvider = {
+    val stsOpt = for {
+      roleArn <- Option(conf.getTrimmed(S3aConstants.ASSUMED_ROLE_ARN))
+      roleSessionName <- Option(conf.getTrimmed(S3aConstants.ASSUMED_ROLE_SESSION_NAME))
+      durationSeconds <- Option(conf.getTimeDuration(S3aConstants.ASSUMED_ROLE_SESSION_DURATION, 0L, TimeUnit.SECONDS).toInt)
+      externalId <- Option(conf.getTrimmed("fs.s3a.assumed.role.session.external.id"))
+    } yield StsAssumeRoleCredentialsProvider.builder
+      .stsClient {
+        StsClient.builder.defaultsMode(DefaultsMode.AUTO).build
```
peel, line 69: I haven't used the `AUTO` mode before. Is it something we've tested? They suggest it's experimental, but I guess it's safe. Just wondered whether you've considered it.

peel (approved):

modules/aws/src/main/scala/com.snowplowanalytics.snowplow.lakes/AssumedRoleCredentialsProvider.scala
istreeter: When working on common-streams, we decided to trust the defaults mode on the kinesis source/sink.  That PR is [here](https://github.com/snowplow-incubator/common-streams/pull/15) but unfortunately the conversation must have happened somewhere private.

The benefits are.... no need to litter our config files with retry settings for each sdk call.  No need to implement the retry-handling.  Our code can trust the sdk about what error classes are re-tryable.

>  Is it something we've tested?

No.  To find out whether this is a good idea or not... I think we need to just get it into production on some low-priority apps, and see if proves to be stable over long time periods (many months).

> They suggest it's experimental,

Out of interest, where did you read that suggestion?  I was following [this blog post](https://aws.amazon.com/blogs/developer/introducing-smart-configuration-defaults-in-the-aws-sdk-for-java-v2/) and [documentation such as this](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/awscore/client/builder/AwsClientBuilder.html#defaultsMode(software.amazon.awssdk.awscore.defaultsmode.DefaultsMode)).

peel: Understood. I was wondering about the choice. This clarifies this. Thanks.

As for why it's experimental: same docs just specifically for [AUTO](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/awscore/defaultsmode/DefaultsMode.html#AUTO) mode.

PR #47, opened 2024-01-27
Slf4j ce3 warning and commit time regression

istreeter:


spenes (approved):
LGTM 👍

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/Environment.scala
```diff
@@ -104,11 +129,26 @@ object Environment {
...
       .toInt
 
+  /**
+   * See the description of `cpuPermit` on the [[Environment]] class
+   *
+   * There must be at least 2 cpu permits, so that the slow spark job (committing to the lake) does
+   * not block the parsing/transformation steps running concurrently.
+   */
+  private def chooseNumCpuPermits(cpuParallelism: Int): Long =
+    Math.max(2L, cpuParallelism.toLong)
```
benjben, line 150: Shouldn't this be `min` ?

PR #50, opened 2024-03-05
Change default log level of Delta libraries

istreeter:
Before this change, the default log level for Delta is WARN. But that means we don't have much visibility in the logs of the granular steps that Delta runs to commit events to the lake.

This change bumps the log level to INFO for Delta only (other logger's are left un-changed).  This adds a modest number of extra lines to the log output.  Not overwhelming, but enough to be helpful.


colmsnowplow (commented):
LGTM

PR #53, opened 2024-03-11
Remove Snowflake as an Iceberg catalog

istreeter:
This removes some placeholder code, which had never been turned into a proper feature.

Previously, we were planning to support loading into Iceberg with Snowflake as the Iceberg catalog.  But now it transpires that the open source Iceberg libraries support only *read* access to Snowflake Iceberg tables. Which is not helpful for this loader.

We can better serve Snowflake users by adding Glue as an Iceberg catalog. Snowflake has support for connecting to Iceberg tables in Glue. It was on our roadmap anyway to add support for Glue.


spenes (approved):
LGTM 👍

PR #55, opened 2024-03-15
Add AWS Glue as an Iceberg catalog

istreeter:
This enables using Redshift or Athena to query Snowplow events as an external Iceberg table in S3.


modules/core/src/main/resources/reference.conf
```diff
@@ -42,6 +42,11 @@
         "hoodie.metadata.index.column.stats.column.list": "load_tstamp,collector_tstamp,derived_tstamp,dvce_created_tstamp"
         // "hoodie.embed.timeline.server.reuse.enabled": "true" // TODO: Experiment with this.
       }
+
+      "catalog": {
+        "type": "Hadoop"
+        "options" {}
```
oguzhanunlu, line 48: `"options": {}` ?

oguzhanunlu (approved):
thanks @istreeter !

modules/core/src/main/resources/reference.conf
istreeter: Whoops, thanks.  Surprisingly, that's actually valid hocon syntax!  But I should change it anyway.

oguzhanunlu: I wasn't 100% sure but good to learn, thanks :D

PR #56, opened 2024-03-19
Bump common-streams to 0.5.0

istreeter:


config/config.aws.reference.hocon
```diff
@@ -138,6 +145,14 @@
     }
   }
 
+  # -- Schemas that won't be loade to the lake.  Optional, default value []
```
pondzix, line 148: Typo! `loade` ;)

PR #59, opened 2024-05-20
Bump spark and lake libraries to latest versions

istreeter:
- Delta bumped to 3.2.0
- Iceberg bumped to 1.5.1
- Spark bumped to 3.5.1

The latest version of Hudi depends on Spark 3.4. So this commit also changes the sbt build so we can use different Spark versions in the different final assets.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,7 +32,7 @@ lazy val core: Project = project
 lazy val azure: Project = project
   .in(file("modules/azure"))
   .settings(BuildSettings.azureSettings)
-  .settings(libraryDependencies ++= Dependencies.azureDependencies35 ++ Dependencies.spark35RuntimeDependencies)
+  .settings(libraryDependencies ++= Dependencies.azureDependencies ++ Dependencies.spark35RuntimeDependencies)
   .dependsOn(core)
   .enablePlugins(BuildInfoPlugin, JavaAppPackaging, SnowplowDockerPlugin)
 
@@ -46,7 +46,7 @@ lazy val gcp: Project = project
 lazy val aws: Project = project
   .in(file("modules/aws"))
   .settings(BuildSettings.awsSettings)
-  .settings(libraryDependencies ++= Dependencies.awsDependencies35 ++ Dependencies.spark35RuntimeDependencies)
+  .settings(libraryDependencies ++= Dependencies.awsDependencies ++ Dependencies.spark35RuntimeDependencies)
   .dependsOn(core)
   .enablePlugins(BuildInfoPlugin, JavaAppPackaging, SnowplowDockerPlugin)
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -32,21 +32,21 @@ lazy val core: Project = project
 lazy val azure: Project = project
   .in(file("modules/azure"))
   .settings(BuildSettings.azureSettings)
-  .settings(libraryDependencies ++= Dependencies.azureDependencies)
+  .settings(libraryDependencies ++= Dependencies.azureDependencies35 ++ Dependencies.spark35RuntimeDependencies)
```
pondzix, line 35: It looks like `azureDependencies35` already brings `sparkCore35`. Why do then also have to provide `spark35RuntimeDependencies` with another `sparkCode35`?

I think I don't understand why now `${cloud}Dependencies` contains spark core dependency. Do we have to do that?

build.sbt
```diff
@@ -32,21 +32,21 @@ lazy val core: Project = project
...
   .dependsOn(core)
   .enablePlugins(BuildInfoPlugin, JavaAppPackaging, SnowplowDockerPlugin)
 
 lazy val gcp: Project = project
   .in(file("modules/gcp"))
   .settings(BuildSettings.gcpSettings)
-  .settings(libraryDependencies ++= Dependencies.gcpDependencies)
+  .settings(libraryDependencies ++= Dependencies.gcpDependencies ++ Dependencies.spark35RuntimeDependencies)
```
pondzix, line 42: Why we have aws/azure 34/35 and not gcp done like that?

build.sbt
istreeter: The azure app needs to [import from hadoop at compile time](https://github.com/snowplow-incubator/snowplow-lake-loader/blob/0.4.0-rc1/modules/azure/src/main/scala/com.snowplowanalytics.snowplow.lakes/AzureTokenProvider.scala#L18-L19).  And the easiest way to do that was to add Spark as a compile time dependency.  I suppose I could have added hadoop to `azureDependencies` instead of adding spark.  Hmmm maybe that's better.

Similarly the aws app needs to [import from hadoop at compile time](https://github.com/snowplow-incubator/snowplow-lake-loader/blob/0.4.0-rc1/modules/aws/src/main/scala/com.snowplowanalytics.snowplow.lakes/AssumedRoleCredentialsProvider.scala#L18-L19).

It's also in `spark35RuntimeDependencies` because that's the only way it gets provided to the gcp app.  The gcp module does not import from spark or hadoop at compile time.

I suppose I could have done this differently.  Either....

1. I could have added hadoop to `azureDependencies` and `awsDependencies` instead of adding all of spark. This would have made it more clear what is needed at compile time.
2. Or, I could have added spark to `gcpDependencies` so at least all clouds look more similar.

Hmmmm maybe 1 is nicer.  What do you think?

build.sbt
istreeter: The gcp app does not need spark/hadoop at compile time (see other comment)

build.sbt
pondzix: I'd go with 1. And I think `awsDependencies` already declares `hadoopAws` (similar for azure and `hadoopAzure`) dependency for compile, so we have everything we need with hadoop for each cloud, right?

pondzix (approved):

PR #60, opened 2024-05-20
Optimizations for larger compute instances

istreeter:
**Increase number of spark writer tasks**. Previously, we did `.coalesce(1)` on the spark data frame, which meant we could write really big parquet files. However, on bigger compute instances the data frame grows so large that writing to storage can become the bottle neck. After this PR, we do `.coalesce(n)` where `n` scales with the number of cpu.

**Changes in how windows overlap**. Finalizing a windows comprises two steps: 1. Union together the pending Data Frames, and 2. write into the Lake. Previously, we did not start on step 1 until the previous window had completely finished step 2. After this PR, we start eagerly on step 1 even if the previous window is still on step 2. This behaviour change is made possible by upgrading common-streams to 0.7.0

**Configurable number of eager windows**. Previously, we let a window begin processing even when the immediate previous windows was still finalizing. Now, there is a configuration option so a window can start even when >1 previous windows are still finalizing. The default value is to retain the previous behaviour. This is an advanced option that probably should not be changed; but it is helpful to make it configurable for some workloads.

**Increased default value of in memory batch size**. Previously, the loader worked with data frames of 25 MB. After the other changes in this PR, I found a value of 50 MB works better on big and small compute instances.


modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Processing.scala
```diff
@@ -61,13 +61,19 @@ object Processing {
...
+  private case class Transformed(
+    events: List[Row],
+    schema: StructType
+  )
+
   private def eventProcessor[F[_]: Async: RegistryLookup](
     env: Environment[F]
   ): EventProcessor[F] = { in =>
     val resources = for {
       now <- Stream.eval(Sync[F].realTime)
```
pondzix, line 73: I think in `WindowState.build` we also get current time to set window start time? Perhaps we could get rid of this one and make this start timestamp field in `WindowState` case class?

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Processing.scala
```diff
@@ -78,13 +84,15 @@ object Processing {
...
-      if (state.framesOnDisk.nonEmpty)
-        env.cpuPermit.surround(env.lakeWriter.removeDataFramesFromDisk(state.framesOnDisk))
-      else
-        Sync[F].unit
-    }
+  /**
+   * Manages the lifecycle of initializing and cleaning a Spark DataFrame scoped to the lifetime of
+   * the window
+   */
+  private def manageDataFrame[F[_]](env: Environment[F], viewName: String): Stream[F, Unit] = {
```
pondzix, line 91: Another idea: how about `WindowState.build` returns `Resource` and we manage init/close there? Then simple `Stream.resource`, maybe this would a bit more explicit what's our window lifecycle?

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkUtils.scala
```diff
@@ -59,44 +58,48 @@ private[processing] object SparkUtils {
...
-    val count = rows.size
+  ): F[Unit] =
     for {
-      viewName <- Sync[F].delay("v" + UUID.randomUUID.toString.replaceAll("-", ""))
-      _ <- Logger[F].debug(s"Saving batch of $count events to local disk")
+      _ <- Logger[F].debug(s"Saving batch of ${rows.size} events to local DataFrame $viewName")
       _ <- Sync[F].blocking {
              spark
                .createDataFrame(rows.toList.asJava, schema)
                .coalesce(1)
```
pondzix, line 80: This `coalesce` is still 1 and it's ok, right?

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkUtils.scala
```diff
@@ -59,44 +58,48 @@ private[processing] object SparkUtils {
...
                .localCheckpoint()
-               .createTempView(viewName)
+               .unionByName(spark.table(viewName), allowMissingColumns = true)
+               .createOrReplaceTempView(viewName)
```
pondzix, line 83: Curious: why `createOrReplaceTempView` here if we do that already in `initializeLocalDataFrame`?

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Processing.scala
```diff
@@ -237,21 +235,18 @@ object Processing {
...
-      }
+      val commit = if (state.numEvents > 0) {
+        for {
+          _ <- Logger[F].info(s"Window ${state.viewName} ready to write and commit ${state.numEvents} events to the lake.")
+          _ <- Logger[F].info(s"Non atomic columns: [${state.nonAtomicColumnNames.toSeq.sorted.mkString(",")}]")
+          _ <- env.lakeWriter.commit(state.viewName)
+          now <- Sync[F].realTime
+          _ <- Logger[F].info(s"Window ${state.viewName} finished writing and committing ${state.numEvents} events to the lake.")
+          _ <- env.metrics.addCommitted(state.numEvents)
+          _ <- env.metrics.setProcessingLatency(now - realTimeWindowStarted)
```
pondzix, line 246: I know this line hasn't changed, but is it OK that we set processing latency only when there are some good events? Maybe that should be separate `Pipe` in `processBatches` where we set metrics regardless of event counts?

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Processing.scala
```diff
@@ -97,46 +105,41 @@ object Processing {
       .through(parseBytes(env, badProcessor))
       .through(handleParseFailures(env))
       .through(BatchUp.noTimeout(env.inMemBatchBytes))
-      .through(sinkParsedBatch(env, badProcessor, ref))
+      .through(transformBatch(env, badProcessor, ref))
+      .through(sinkTransformedBatch(env, ref))
```
pondzix, line 109: Curious: any specific reason why `finalizeWindow` is not here as another `Pipe` but it's `append`ed later instead?

istreeter: In my mind, `processBatches` is "all the stuff we do for every single batch".  And `finalizeWindow` stands apart because it is something we do once per window.

Syntactically you are right I could add it here like....

```scala
  ...
  .through(transformBatch(env, badProcessor, ref))
  .through(sinkTransformedBatch(env, ref))
  .through(stream => stream.append(finalizeWindow(env, stateRef)))
```

I liked the separation.  But I'd also change it if you like the other way.

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Processing.scala
istreeter: Processing latency is interesting because it is only defined after writing some events.  For example, imagine we have a windowing of 5 minutes, so we only commit events every 5 minutes.  And imagine we emit metrics every 1 minute.  The processing metric over time looks like this:

```
12:01:00 undefined
12:02:00 undefined
12:03:00 undefined
12:04:00 undefined
12:05:00 320 seconds
12:06:00 undefined
12:07:00 undefined
12:08:00 undefined
12:09:00 undefined
12:10:00 340 seconds
```

So... sure we could emit a zero instead of undefined.  But it doesn't really mean anything, and it is even misleading.  E.g. at 12:07:00 is the processing latency really zero?

This metric mainly exists to help us developers understand how the lake loader is behaving.  So we are free to make the metric behave however we find helpful.

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkUtils.scala
istreeter: Yes here it is ok.  It means the individual small dataframes, which don't need to be split any further.

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkUtils.scala
istreeter: Think of this as `.replaceTempView` (but Spark doesn't have a method called that).

Data Frames are immutable.  So the loader runs like this:

- Create an empty DataFrame and save it under the name `v123`
- Later, create a new DataFrame by union'ing `v123` with some new data.  Save that new DataFrame under the name `v123`
- Same again
- etc

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Processing.scala
pondzix: Yes, emitting zeros is not useful. But I thought more about bad data. Let's say for some reason loader produces only badrows, no good data. Then we don't have this processing latency set, right? So loader has done some processing, but its latency is not reported. Just a thought, not sure that would be really useful 🤷

pondzix (approved):

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Processing.scala
istreeter: Oh I see what you mean now.

Hmmm... probably the same argument still applies.  If there is no good data for an entire window, then "processing latency" is not well defined.  I cannot think how I would benefit from seeing processing latency if all data went to bad.

If we cared about latency of data going to bad, then it should be a different metric and we should emit it every time there is bad data.  Currently, though, I don't care about the latency of data going to bad.

PR #66, opened 2024-06-20
Allow disregarding Iglu field's nullability when creating output columns

istreeter:
Some query engines dislike Iceberg tables in which a STRUCT field is nullable and a nested field of the STRUCT is non-nullable. For example, in Snowflake we have seen errors like "SQL execution internal error" when the nested field contains a null.

This PR adds a config option `respectNullability`.  When set to `false`, the Lake Loader will make all nested fields nullable. This is slightly less ideal for data storage, but it brings back compatibility with query engines like Snowflake.

The default value of the new config option is `true` which maintains the behaviour of previous versions of the loader.

### DIFF
--- a/config/config.aws.reference.hocon
+++ b/config/config.aws.reference.hocon
@@ -190,8 +190,10 @@
   ]
 
   # -- Whether the output parquet files should declare nested fields as non-nullable according to the Iglu schema.
+  # -- When true (default), nested fields are nullable only if they are not required fields according to the Iglu schema.
+  # -- When false, all nested fields are defined as nullable in the output table's schemas
   # -- Set this to false if you use a query engine that dislikes non-nullable nested fields of a nullable struct.
-  "respectNullability": true
+  "respectIgluNullability": true
 
   "monitoring": {
     "metrics": {
--- a/config/config.aws.reference.hocon
+++ b/config/config.aws.reference.hocon
@@ -189,6 +189,10 @@
     "iglu:com.acme/skipped4/jsonschema/*-*-*"
   ]
 
+  # -- Whether the output parquet files should declare nested fields as non-nullable according to the Iglu schema.
+  # -- Set this to false if you use a query engine that dislikes non-nullable nested fields of a nullable struct.
+  "respectNullability": true
+
   "monitoring": {
     "metrics": {
 
--- a/modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkSchema.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkSchema.scala
@@ -22,11 +22,11 @@ object SparkSchema {
    *
    * The returned schema includes atomic fields and non-atomic fields but not the load_tstamp column
    */
-  private[processing] def forBatch(entities: Vector[TypedTabledEntity], respectNullability: Boolean): StructType = {
+  private[processing] def forBatch(entities: Vector[TypedTabledEntity], respectIgluNullability: Boolean): StructType = {
     val nonAtomicFields = entities.flatMap { tte =>
       tte.mergedField :: tte.recoveries.map(_._2)
     }
-    StructType(atomic ++ nonAtomicFields.map(asSparkField(_, respectNullability)))
+    StructType(atomic ++ nonAtomicFields.map(asSparkField(_, respectIgluNullability)))
   }
 
   /**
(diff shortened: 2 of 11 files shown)


config/config.aws.reference.hocon
```diff
@@ -189,6 +189,10 @@
     "iglu:com.acme/skipped4/jsonschema/*-*-*"
   ]
 
+  # -- Whether the output parquet files should declare nested fields as non-nullable according to the Iglu schema.
+  # -- Set this to false if you use a query engine that dislikes non-nullable nested fields of a nullable struct.
+  "respectNullability": true
```
istreeter, line 194: Whoever reviews this, please comment on the name of this new config option.  And whether the description of this option is helpful.

The new config option will be documented on the snowplow docs site, but the description will probably just be copied over from this example config file.

pondzix: Do I understand it correctly:

* when `true` - use nullability definition from Iglu Schema
* when `false` - always makes nested fields nullable
?

Maybe we could include such `when true/false, then....` in description? Also about this fragment:

`Whether the output parquet files should declare nested fields as non-nullable according to the Iglu schema...`

it suggests that for `true` nested fields are always non-nullable, but they still can be nullable if this is what we have in Iglu schema...?

Another ideas for variants of the setting's name:

* `forceNullability`
* `alwaysNullable`
* `respectSchemaNullability`
*  `respectIgluNullability`

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkSchema.scala
```diff
@@ -37,29 +37,30 @@ object SparkSchema {
    * @note
    *   this is a `val` not a `def` because we use it over and over again.
    */
-  val atomic: Vector[StructField] = AtomicFields.static.map(asSparkField)
+  val atomic: Vector[StructField] = AtomicFields.static.map(asSparkField(_, true))
```
pondzix, line 40: Does it also affect top-level unstruct/contexts fields and not only nested fields?

If so, is it OK? I guess here we care only about nested fields?

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkSchema.scala
```diff
@@ -22,11 +22,11 @@ object SparkSchema {
    *
    * The returned schema includes atomic fields and non-atomic fields but not the load_tstamp column
    */
-  private[processing] def forBatch(entities: Vector[TypedTabledEntity]): StructType = {
+  private[processing] def forBatch(entities: Vector[TypedTabledEntity], respectNullability: Boolean): StructType = {
     val nonAtomicFields = entities.flatMap { tte =>
       tte.mergedField :: tte.recoveries.map(_._2)
     }
-    StructType(atomic ++ nonAtomicFields.map(asSparkField))
+    StructType(atomic ++ nonAtomicFields.map(asSparkField(_, respectNullability)))
```
pondzix, line 29: Does it also affect top-level unstruct/contexts fields and not only nested fields?

If so, is it OK? I guess here we care only about nested fields?

istreeter: I think I've caused confusion about what I mean about a nested field.  If a schema has fields `abc` and `xyz`, then I'm referring to `unstruct_event_myvendor_myevent_1.abc` and `unstruct_event_myvendor_myevent_1.xyz` as nested fields.  Sure they are top level fields of the Iglu schema, but they are nested fields from the point of view of the events table.

Does that resolve your question, or have I just made it even more confusing?

pondzix: And I think I've cause confusion about top level fields 😁 

What I meant was not even `abc` field but rather whole parent `unstruct_event_myvendor_myevent_1` field (or `contexts_myvendor_myevent_1`), columns living next to atomic. These are also fields created by schema-ddl and they also have some nullability (both are nullable...?) It doesn't probably really matter.

pondzix (approved):

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkSchema.scala
istreeter: OK I see what you mean.

This whole PR was about taking columns which were previously non-nullable and making them nullable.  But the parent fields (e.g. `unstruct_event_myvendor_myevent_1`) were always nullable and always will be.  So no change is needed there.

Or the other way of answering.... relax, this works.  We tested it.

PR #69, opened 2024-07-12
Extend health probe to report unhealthy on more error scenarios

istreeter:
This PR follows the same style of health probe we did in [BigQuery Loader v2](https://github.com/snowplow-incubator/snowplow-bigquery-loader/blob/c1951fe344ce2fd44e1dab9e48499d9d08224859/modules/core/src/main/scala/com.snowplowanalytics.snowplow.bigquery/AppHealth.scala) and [in Snowflake Loader](https://github.com/snowplow-incubator/snowflake-loader/blob/0.2.3/modules/core/src/main/scala/com.snowplowanalytics.snowplow.snowflake/AppHealth.scala).

It adds a new trait `LakeWriter.WithHandledErrors` which similar to [what we did here in BigQuery Loader](https://github.com/snowplow-incubator/snowplow-bigquery-loader/blob/c1951fe344ce2fd44e1dab9e48499d9d08224859/modules/core/src/main/scala/com.snowplowanalytics.snowplow.bigquery/processing/TableManager.scala#L58).  It means we can use the same nice pattern of writing unit tests for this class.  In follow up work we can extend `WithHandledErrors` to manage the retry logic and reporting setup errors to a webhook.

### DIFF
--- a/modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriterSpec.scala
+++ b/modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriterSpec.scala
@@ -1,0 +1,221 @@
+/*
+ * Copyright (c) 2014-present Snowplow Analytics Ltd. All rights reserved.
+ *
+ * This software is made available by Snowplow Analytics, Ltd.,
+ * under the terms of the Snowplow Limited Use License Agreement, Version 1.0
+ * located at https://docs.snowplow.io/limited-use-license-1.0
+ * BY INSTALLING, DOWNLOADING, ACCESSING, USING OR DISTRIBUTING ANY PORTION
+ * OF THE SOFTWARE, YOU AGREE TO THE TERMS OF SUCH LICENSE AGREEMENT.
+ */
+
+package com.snowplowanalytics.snowplow.lakes.processing
+
+import cats.implicits._
+import cats.data.NonEmptyList
+import cats.effect.{IO, Ref}
+import org.specs2.Specification
+import cats.effect.testing.specs2.CatsEffect
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.Row
+
+import com.snowplowanalytics.snowplow.runtime.HealthProbe
+import com.snowplowanalytics.snowplow.lakes.AppHealth
+import com.snowplowanalytics.snowplow.sources.{EventProcessingConfig, EventProcessor, SourceAndAck}
+
+import scala.concurrent.duration.{DurationLong, FiniteDuration}
+
+class LakeWriterSpec extends Specification with CatsEffect {
+  import LakeWriterSpec._
+
+  def is = s2"""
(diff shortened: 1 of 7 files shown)


modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriterSpec.scala
```diff
@@ -0,0 +1,221 @@
...
+    become unhealthty after failure to commit to the lake $e4
+  """
+
+  def e1 =
+    control().flatMap { c =>
+      val expected = Vector(
+        Action.CreateTableAttempted
+      )
+
+      val wrappedLakeWriter = LakeWriter.withHandledErrors(c.lakeWriter, c.appHealth)
```
pondzix, line 44: As I think this is common for each test, it means it can be part of `Control`?

modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriterSpec.scala
```diff
@@ -0,0 +1,221 @@
...
+
+class LakeWriterSpec extends Specification with CatsEffect {
+  import LakeWriterSpec._
+
+  def is = s2"""
+  The lake writer should:
+    become healthy after creating the table $e1
+    stay unhealthy after failure to create the table $e2
+    stay healthy after committing to the lake $e3
+    become unhealthty after failure to commit to the lake $e4
```
pondzix, line 35:
```suggestion
    become unhealthy after failure to commit to the lake $e4
```

modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriterSpec.scala
```diff
@@ -0,0 +1,221 @@
...
+import scala.concurrent.duration.{DurationLong, FiniteDuration}
+
+class LakeWriterSpec extends Specification with CatsEffect {
+  import LakeWriterSpec._
+
+  def is = s2"""
+  The lake writer should:
+    become healthy after creating the table $e1
+    stay unhealthy after failure to create the table $e2
+    stay healthy after committing to the lake $e3
```
pondzix, line 34: Stay or become in this e3 case? 😁

modules/core/src/test/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriterSpec.scala
istreeter: It could be part of `Control`.

In these tests, the thing we are testing is the behaviour of `withHandledErrors`.  So I quite like the pattern of explicitly calling the function under test, in each spec.

This is a matter-of-opinion thing.  If you ask me again next month I might say the opposite.

pondzix (approved):

PR #70, opened 2024-07-15
Bump hudi to 0.15.0

istreeter:
As part of this change we also bump spark to 3.5.x when using Hudi. And bump scala to 2.13.x. Previously we were pinned to earlier versions because of compatibility with Hudi 0.14.0.

This PR is implemented in a way that we retain the flexibility of easily supporting a different version of Spark for the Hudi docker image. I anticipate we might need this flexibility if Iceberg/Delta are faster to add support for Spark 4.x.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,21 +32,21 @@ lazy val core: Project = project
 lazy val azure: Project = project
   .in(file("modules/azure"))
   .settings(BuildSettings.azureSettings)
-  .settings(libraryDependencies ++= Dependencies.azureDependencies ++ Dependencies.spark35RuntimeDependencies)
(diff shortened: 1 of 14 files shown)


istreeter:
This PR also disables the HudiSpec.  This is an unfortunate necessity, because since Hudi 0.15.0 it is not possible to use Spark to read partition column values if `TimestampBasedKeyGenerator` is used.  This will be fixed in Hudi 1.0.0, see apache/hudi#11615.

This issue only affects _reading_ Hudi, so it does not affect the ability of this loader to write Hudi using version 0.15.0.  Until HudiSpec is re-enabled, we need to use other means to check that this loader can write Hudi correctly.

benjben (approved):
Nice :ok_hand:

PR #71, opened 2024-07-20
Bump aws-hudi to 1.0.0-beta2

istreeter:
Although we remain on Hudi 0.15.0, this bumps the AWS module of Hudi to version 1.x. This is safe because the AWS module is backwards compatible to 0.15.0. This change lets us use features only in 1.x like assumed role credentials provider.


colmsnowplow (approved):
🙌

PR #73, opened 2024-07-24
Implement alerting and retrying mechanisms

spenes:
Jira ref: PDP-1321

Two features has been added in this PR: alerting and retrying

For alerting, webhook method is used similar to other Snowplow apps. Alert message is sent to URL given in the config. Alerts are sent for some error cases, not for all of them. In this PR, it is implemented such that it is sent only for setup errors. The error cases where alert sent can be extended in the future, of course. 

For retrying, two retry policies can be defined [similar to Snowflake Loader](https://github.com/snowplow-incubator/snowflake-loader/blob/main/modules/core/src/main/scala/com.snowplowanalytics.snowplow.snowflake/processing/SnowflakeRetrying.scala). One of them is for setup errors and other one is for transient errors. Alert would be sent only for setup errors, not for transient errors. 

Also, possible setup error cases for Iceberg/Glue/S3 are added in this PR as well. Error cases for other destinations/table formats will be added later.

One other thing I want to mention is changes on the dependencies and SBT packages. Some of the dependencies are switched from runtime to compile-time. This is done to be able to use exception classes in those dependencies in the project while trying to determine whether thrown exception is for one of the setup error cases or not. Also, new `iceberg` package is created to be able to use Iceberg specific exception classes while checking setup errors.

### DIFF
--- a/modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/Alert.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/Alert.scala
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2014-present Snowplow Analytics Ltd. All rights reserved.
+ *
+ * This software is made available by Snowplow Analytics, Ltd.,
+ * under the terms of the Snowplow Limited Use License Agreement, Version 1.0
+ * located at https://docs.snowplow.io/limited-use-license-1.0
+ * BY INSTALLING, DOWNLOADING, ACCESSING, USING OR DISTRIBUTING ANY PORTION
+ * OF THE SOFTWARE, YOU AGREE TO THE TERMS OF SUCH LICENSE AGREEMENT.
+ */
+
+package com.snowplowanalytics.snowplow.lakes
+
+import cats.Show
+import cats.implicits.showInterpolator
+
+import com.snowplowanalytics.iglu.core.circe.implicits.igluNormalizeDataJson
+import com.snowplowanalytics.iglu.core.{SchemaKey, SchemaVer, SelfDescribingData}
+import com.snowplowanalytics.snowplow.runtime.AppInfo
+
+import io.circe.Json
+import io.circe.syntax.EncoderOps
+
+import java.sql.SQLException
+
+sealed trait Alert
+object Alert {
+
+  /** Restrict the length of an alert message to be compliant with alert iglu schema */
+  private val MaxAlertPayloadLength = 4096
+
+  final case class FailedToCreateEventsTable(cause: Throwable) extends Alert
+
+  def toSelfDescribingJson(
+    alert: Alert,
+    appInfo: AppInfo,
+    tags: Map[String, String]
+  ): Json =
+    SelfDescribingData(
+      schema = SchemaKey("com.snowplowanalytics.monitoring.loader", "alert", "jsonschema", SchemaVer.Full(1, 0, 0)),
+      data = Json.obj(
+        "appName" -> appInfo.name.asJson,
+        "appVersion" -> appInfo.version.asJson,
+        "message" -> getMessage(alert).asJson,
+        "tags" -> tags.asJson
+      )
+    ).normalize
+
+  private def getMessage(alert: Alert): String = {
+    val full = alert match {
+      case FailedToCreateEventsTable(cause) => show"Failed to create events table: $cause"
+    }
+
+    full.take(MaxAlertPayloadLength)
+  }
+
+  private implicit def throwableShow: Show[Throwable] = {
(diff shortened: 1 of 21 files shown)


packaging/iceberg/src/main/scala/com.snowplowanalytics.snowplow.lakes/TableFormatSetupError.scala
```diff
@@ -0,0 +1,29 @@
...
+package com.snowplowanalytics.snowplow.lakes
+
+import org.apache.iceberg.exceptions.{ForbiddenException => IcebergForbiddenException, NotFoundException => IcebergNotFoundException}
+
+object TableFormatSetupError {
+
+  // Check if given exception is specific to iceberg format
+  def check(t: Throwable): Boolean =
+    t match {
+      case _: IcebergNotFoundException =>
```
istreeter, line 20: Have you considered in which package/class we are going to match for Delta setup errors?

If it's going to be in _this_ package, then it's better to call this package something like `packaging/delta-iceberg` instead of `packaging/iceberg`.

On the other hand, if we need another package for Delta.... then I don't know how it's going to work to have two packages both providing a `TableFormatSetupError` object.

Or maybe you haven't thought that far ahead!!  That's fine.  We can tackle that problem when we work on the Delta setup errors.

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/Alert.scala
```diff
@@ -0,0 +1,81 @@
...
+import com.snowplowanalytics.iglu.core.{SchemaKey, SchemaVer, SelfDescribingData}
+import com.snowplowanalytics.snowplow.runtime.AppInfo
+
+import io.circe.Json
+import io.circe.syntax.EncoderOps
+
+import java.sql.SQLException
+
+sealed trait Alert
+object Alert {
```
istreeter, line 26: There's a lot of stuff in this class I think we could move to common-streams.  It is near identical to Alert in the other loaders.

That's beyond the scope of this current PR, but I'll try to remember it as some tidy-up we should do later.

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/Monitoring.scala
```diff
@@ -0,0 +1,76 @@
...
+import org.http4s.{EntityDecoder, Method, Request}
+
+import org.typelevel.log4cats.Logger
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+trait Monitoring[F[_]] {
+  def alert(message: Alert): F[Unit]
+}
+
+object Monitoring {
```
istreeter, line 29: Maybe this could also move to common-streams in follow-up work.

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/Retrying.scala
```diff
@@ -0,0 +1,84 @@
...
+      )
+      .retryingOnAllErrors(
+        policy  = policyForTransientErrors[F](config),
+        onError = logError[F](_, _)
+      )
+
+  private def policyForSetupErrors[F[_]: Applicative](config: Config.Retries): RetryPolicy[F] =
+    RetryPolicies.exponentialBackoff[F](config.setupErrors.delay)
+
+  private def policyForTransientErrors[F[_]: Applicative](config: Config.Retries): RetryPolicy[F] =
```
istreeter, line 64: Spark already does retries, and we already make it configurable [here](https://github.com/snowplow-incubator/snowplow-lake-loader/blob/0.4.1/config/config.aws.reference.hocon#L165-L166).  So I don't know how the spark retries and "transient error" retries will interact with each other.

Let's leave it like this though, because I don't want to remove either retry!  (until the day we see some retry behaviour that we don't like)

istreeter (commented):
I bet the hardest part was working on `build.sbt` and `Dependencies.scala`?

Dependency management is a bit awkward in this app -- you probably realized!  It's partly so we can support different versions of Spark for different builds, although we're not doing that at the moment.

packaging/iceberg/src/main/scala/com.snowplowanalytics.snowplow.lakes/TableFormatSetupError.scala
spenes: This package is created purely for dependency management. Otherwise, we would have to include Iceberg dependencies to Hudi artifact. It seems like delta-iceberg dependencies are always coming together therefore I don't think we need separate package/class for Delta setup errors. I will change its name to  `packaging/delta-iceberg` as you suggested 👍

istreeter (approved):

PR #74, opened 2024-07-31
Add alert & retry for delta/s3 initialization

oguzhanunlu:
ref: PDP-1322

### DIFF
--- a/modules/aws/src/main/scala/com.snowplowanalytics.snowplow.lakes/AwsApp.scala
+++ b/modules/aws/src/main/scala/com.snowplowanalytics.snowplow.lakes/AwsApp.scala
@@ -44,7 +44,7 @@ object AwsApp extends LoaderApp[KinesisSourceConfig, KinesisSinkConfig](BuildInf
       // no such bucket exist
       true
     case _: AccessDeniedException =>
-      // 1 - no permission to see s3 bucket
+      // 1 - s3 bucket's permission policy denies all actions
       // 2 - not authorized to assume the role
       true
     case t => TableFormatSetupError.check(t)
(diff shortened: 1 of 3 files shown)


modules/aws/src/main/scala/com.snowplowanalytics.snowplow.lakes/AwsApp.scala
```diff
@@ -36,6 +40,13 @@ object AwsApp extends LoaderApp[KinesisSourceConfig, KinesisSinkConfig](BuildInf
     case _: StsException =>
       // No permission to assume the role given to authenticate to S3/Glue
       true
+    case _: UnknownStoreException =>
+      // no such bucket exist
+      true
+    case _: AccessDeniedException =>
+      // 1 - no permission to see s3 bucket
+      // 2 - not authorized to assume the role
+      true
```
istreeter, line 49: Please can you explain to me under which circumstances you get a `AccessDeniedException`?

Can you confirm I've understood your code comments correctly....

- If the bucket doesn't exist, or we don't have permission to see the bucket itself --> get a `UnknownStoreException`
- If we don't have access to read/write objects inside the bucket --> get a `DeltaIOException`

...so I can't work out what scenario gives a `AccessDeniedException`.

oguzhanunlu: > I can't work out what scenario gives a AccessDeniedException.

we get `AccessDeniedException` in these cases:

- when the bucket's policy denies all actions, we get AccessDeniedException without any error message
- when we try to assume a role that doesn't exist, we get AccessDeniedException with detailed error message

I see that my comment on L47 is misleading

oguzhanunlu: just pushed a commit to describe exact situation

istreeter (approved):

PR #75, opened 2024-08-05
Make alert messages more human-readable

istreeter:
The webhook alert should contain a short helpful message explaining why an error is caused by the destination setup.  In other snowplow loaders we get the message simply by serializing the Exception.  But in Lake Loader I found the exception messages to be very messy.

In a related problem, for Hudi setup errors I needed to traverse the Exception's `getCause` in order to check if it was a setup error.

This PR takes more explicit control of setting short friendly error messages, and traversing the `getCause` to get all relevant messages.

E.g. an alert message before this change:

> Failed to create events table: s3a://<REDACTED/events/_delta_log: org.apache.hadoop.fs.s3a.auth.NoAuthWithAWSException: No AWS Credentials provided by V1ToV2AwsCredentialProviderAdapter : software.amazon.awssdk.services.sts.model.StsException: User: arn:aws:iam::<REDACTED>:user/<REDACTED> is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::<REDACTED>:role/<REDACTED> (Service: Sts, Status Code: 403, Request ID: 00000000-0000-0000-0000-000000000000)

The corresponding alert after this change:

> Failed to create events table: s3a://<REDACTED/events/_delta_log: Failed to initialize AWS access credentials: Missing permissions to assume the AWS IAM role

**Other small changes I snuck into this commit:**

- Added specific webhook alerts for Hudi.
- Removed the AssumedRoleCredentialsProvider for aws sdk v1.  This is no longer needed now that Hadoop is fully using aws sdk v2.
- Fixed minor bug with retrying creating a database in Hudi Writer


oguzhanunlu (approved):
👍 thanks @istreeter !

PR #76, opened 2024-08-06
Iceberg fail fast if missing permissions on the catalog

istreeter:
The loader has a feature where it sends alerts to a webhook if the destination is mis-configured.

However, with Iceberg it was possible for the initialization to be successful, and yet the loader might still fail later while committing events, e.g. due to a permissions error updating the Glue catalog.

Here we add an extra step so the loader is forced to make an empty commit early during initialization.  If there is an error committing, then the loader sends a webhook alert and never becomes healthy.


oguzhanunlu (approved):
👍

PR #78, opened 2024-08-16
common-streams 0.8.x with refactored health monitoring

istreeter:
In common-streams 0.8.x we shifted alerting / retrying / webhook out of the applications and into the common library.  It also adds new features like heartbeat webhooks starting when the loader first becomes healthy.

See snowplow-incubator/common-streams#82


modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/Alert.scala
```diff
@@ -13,58 +13,16 @@ package com.snowplowanalytics.snowplow.lakes
...
-    def removeDuplicateMessages(in: List[String]): List[String] =
-      in match {
-        case h :: t :: rest =>
-          if (h.contains(t)) removeDuplicateMessages(h :: rest)
-          else if (t.contains(h)) removeDuplicateMessages(t :: rest)
-          else h :: removeDuplicateMessages(t :: rest)
-        case fewer => fewer
-      }
+  final case class FailedToCreateEventsTable(causes: SetupExceptionMessages) extends Alert
+  final case class FailedToCommitEvents(causes: SetupExceptionMessages) extends Alert
```
pondzix, line 22: Is this one used anywhere?

pondzix (approved):

PR #79, opened 2024-08-30
Add option to exit on missing Iglu schemas

istreeter:
Before this PR, the loader would generate a failed event if it failed to fetch a required schema from Iglu.  However, all events have already passed validation in Enrich, so it is completely unexpected to have an Iglu failure.  An Iglu error _probably_ means some type of configuration error or service outage.

After this PR, the loader will crash and exit on an Iglu error, instead of creating a failed event.  This is probably the preferred behaviour, while the pipeline operator addresses the underlying infrastructure problem.

If an Iglu schema is genuinely now unavailable, then the pipeline operator can override the default behaviour by setting `exitOnMissingFailure: false` in the configuration file or by listing the missing schema in `skipschemas`.

### DIFF
--- a/config/config.aws.reference.hocon
+++ b/config/config.aws.reference.hocon
@@ -207,6 +207,12 @@
     "iglu:com.acme/skipped4/jsonschema/*-*-*"
   ]
 
+  # -- Whether the loader should crash and exit if it fails to resolve an Iglu Schema.
+  # -- We recommend `true` because Snowplow enriched events have already passed validation, so a missing schema normally
+  # -- indicates an error that needs addressing.
+  # -- Change to `false` so events go the failed events stream instead of crashing the loader.
+  "exitOnMissingIgluSchema": true
+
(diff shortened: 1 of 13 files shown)


config/config.aws.reference.hocon
```diff
@@ -207,6 +207,12 @@
     "iglu:com.acme/skipped4/jsonschema/*-*-*"
   ]
 
+  # -- Whether the loader should crash and exit if it fails to resolve an Iglu Schema.
+  # -- We recommend `true` because Snowplow enriched events have already passed validation, so a missing schema normally
+  # -- indicates an error that needs addressing.
+  # -- Change to `false` so events go the failed events stream instead of crashing the loader.
+  "exitOnMissingIgluSchema": true
```
pondzix, line 214: nit: this doesn't match what is in commit message?

pondzix (approved):

config/config.aws.reference.hocon
istreeter: I'll fix the commit message as I merge it.  Thanks.

PR #81, opened 2024-09-09
common-streams 0.8.x with refactored kinesis source

istreeter:
The following improvements are introduced via common-streams 0.8.0-M4:

- Fields starting with a digit are now prefixed with an underscore `_`.  This is needed for Hudi, which does not allow fields starting with a digit (https://github.com/snowplow/schema-ddl/pull/209)
- New kinesis source implementation without fs2-kinesis (https://github.com/snowplow-incubator/common-streams/pull/84)
- Iglu schemas are resolved in parallel, for short pause times during event processing (https://github.com/snowplow-incubator/common-streams/pull/85)
- Common http client configured with restricted max connections per server (https://github.com/snowplow-incubator/common-streams/pull/87)
- Iglu scala client 3.2.0 no longer relies on the "list" schemas endpoint (https://github.com/snowplow/iglu-scala-client/pull/255)

### DIFF
--- a/config/config.aws.reference.hocon
+++ b/config/config.aws.reference.hocon
@@ -34,10 +34,6 @@
       "maxRecords": 1000
     }
 
-    # -- The number of batches of events which are pre-fetched from kinesis.
-    # -- Increasing this above 1 is not known to improve performance.
-    "bufferSize": 3
-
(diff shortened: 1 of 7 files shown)


oguzhanunlu (approved):
💯

PR #88, opened 2024-10-17
PDP-1324 - Handle initialization errors for lake loader Delta Azure

benjben:
Errors handled:
- [x] Wrong OAuth endpoint
- [x] Wrong client id
- [x] Wrong secret
- [x] Wrong container name (doesn't exist)
- [x] Service principal exists but no role assigned to it for the container
- [x] Wrong role assigned to the service principal for the container (e.g. `Storage Blob Data Reader`)
- [x] Wrong storage name

### DIFF
--- a/modules/azure/src/main/scala/com.snowplowanalytics.snowplow.lakes/AuthenticationError.scala
+++ b/modules/azure/src/main/scala/com.snowplowanalytics.snowplow.lakes/AuthenticationError.scala
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2014-present Snowplow Analytics Ltd. All rights reserved.
+ *
+ * This software is made available by Snowplow Analytics, Ltd.,
+ * under the terms of the Snowplow Limited Use License Agreement, Version 1.0
+ * located at https://docs.snowplow.io/limited-use-license-1.0
+ * BY INSTALLING, DOWNLOADING, ACCESSING, USING OR DISTRIBUTING ANY PORTION
+ * OF THE SOFTWARE, YOU AGREE TO THE TERMS OF SUCH LICENSE AGREEMENT.
+ */
+
+package com.snowplowanalytics.snowplow.lakes
+
+import org.apache.hadoop.fs.azurebfs.contracts.exceptions.AbfsRestOperationException
+
+import cats.implicits._
+
+object AuthenticationError {
+
+  private val knownErrors = Map(
+    90002 -> "Tenant is invalid or does not have an active subscription",
+    700016 -> "Application ID was not found for this tenant",
+    7000215 -> "Invalid client secret provided"
+  )
+
+  def unapply(throwable: Throwable): Option[String] = throwable match {
+    case e: AbfsRestOperationException if e.getStatusCode == -1 =>
+      extractErrorCode(e.getMessage).flatMap(knownErrors.get).map(err => s"Could not authenticate. $err")
+    case _ =>
+      None
+  }
+
+  private def extractErrorCode(errorMessage: String): Option[Int] = {
+    val regex = """.*\bAADSTS(\d+)\b.*""".r
+    Either.catchNonFatal {
+      val regex(errorCode) = errorMessage
+      errorCode.toInt
+    }.toOption
+  }
+}
--- a/modules/azure/src/main/scala/com.snowplowanalytics.snowplow.lakes/AzureApp.scala
+++ b/modules/azure/src/main/scala/com.snowplowanalytics.snowplow.lakes/AzureApp.scala
@@ -10,7 +10,11 @@
 
 package com.snowplowanalytics.snowplow.lakes
 
+import java.net.UnknownHostException
+
 import scala.reflect._
+
+import org.apache.hadoop.fs.azurebfs.contracts.exceptions.AbfsRestOperationException
 
 import com.snowplowanalytics.snowplow.sources.kafka.{KafkaSource, KafkaSourceConfig}
 import com.snowplowanalytics.snowplow.sinks.kafka.{KafkaSink, KafkaSinkConfig}
@@ -35,5 +39,23 @@ object AzureApp extends LoaderApp[KafkaSourceConfig, KafkaSinkConfig](BuildInfo)
 
   override def badSink: SinkProvider = KafkaSink.resource(_, classTag[SinkAuthHandler])
 
-  override def isDestinationSetupError: DestinationSetupErrorCheck = TableFormatSetupError.check
+  override def isDestinationSetupError: DestinationSetupErrorCheck = {
+    // Authentication issue (wrong OAuth endpoint, wrong client id, wrong secret)
+    case AuthenticationError(e) =>
+      e
+    // Wrong container name
+    case e: AbfsRestOperationException if e.getStatusCode == 404 =>
+      s"The specified filesystem does not exist (e.g. wrong container name)"
+    // Service principal missing permissions for container (role assignement missing or wrong role)
+    case e: AbfsRestOperationException if e.getStatusCode == 403 =>
+      s"Missing permissions for the destination (needs \"Storage Blob Data Contributor\" assigned to the service principal for the container)"
+    // Soft delete not disabled
+    case e: AbfsRestOperationException if e.getStatusCode == 409 =>
+      "Blob soft delete must be disabled on the storage account"
+    case _: UnknownHostException =>
+      "Wrong storage name"
+    // Exceptions common to the table format - Delta/Iceberg/Hudi
+    case TableFormatSetupError.check(t) =>
+      t
+  }
 }
--- a/modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/tables/DeltaWriter.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/tables/DeltaWriter.scala
@@ -10,6 +10,8 @@
 
 package com.snowplowanalytics.snowplow.lakes.tables
 
+import java.net.InetAddress
+
 import cats.implicits._
 import cats.effect.Sync
 import org.typelevel.log4cats.Logger
@@ -55,7 +57,16 @@ class DeltaWriter(config: Config.Delta) extends Writer {
         .build()
     }: Unit
 
+    // For Azure a wrong storage name means an invalid hostname and infinite retries when creating the Delta table
+    // If the hostname is invalid, UnknownHostException gets thrown
+    val checkHostname =
+      if (List("abfs", "abfss").contains(config.location.getScheme))
+        Sync[F].blocking(InetAddress.getByName(config.location.getHost()))
+      else
+        Sync[F].unit
+
     Logger[F].info(s"Creating Delta table ${config.location} if it does not already exist...") >>
+      checkHostname >>
       Sync[F]
         .blocking(builder.execute())
         .void
--- a/modules/azure/src/test/scala/com.snowplowanalytics.snowplow.lakes/AuthenticationErrorSpec.scala
+++ b/modules/azure/src/test/scala/com.snowplowanalytics.snowplow.lakes/AuthenticationErrorSpec.scala
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2014-present Snowplow Analytics Ltd. All rights reserved.
+ *
+ * This software is made available by Snowplow Analytics, Ltd.,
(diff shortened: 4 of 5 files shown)


modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/tables/DeltaWriter.scala
```diff
@@ -33,6 +35,10 @@ class DeltaWriter(config: Config.Delta) extends Writer {
     )
 
   override def prepareTable[F[_]: Sync](spark: SparkSession): F[Unit] = {
+    val checkHost = Sync[F].blocking {
+      InetAddress.getByName(config.location.getHost())
```
benjben, line 39: The stack trace for a wrong storage name contains
```
[error] Caused by: java.net.UnknownHostException: lbwwlakeloaderaccountfoo.dfs.core.windows.net
[error] 	at java.base/java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:229)
[error] 	at java.base/java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
[error] 	at java.base/java.net.Socket.connect(Socket.java:609)
[error] 	at org.wildfly.openssl.OpenSSLSocket.connect(OpenSSLSocket.java:563)
[error] 	at java.base/sun.net.NetworkClient.doConnect(NetworkClient.java:177)
[error] 	at java.base/sun.net.www.http.HttpClient.openServer(HttpClient.java:509)
[error] 	at java.base/sun.net.www.http.HttpClient.openServer(HttpClient.java:604)
[error] 	at java.base/sun.net.www.protocol.https.HttpsClient.<init>(HttpsClient.java:266)
[error] 	at java.base/sun.net.www.protocol.https.HttpsClient.New(HttpsClient.java:373)
```
so which is also in `java.net`

modules/azure/src/main/scala/com.snowplowanalytics.snowplow.lakes/AzureApp.scala
```diff
@@ -35,5 +39,23 @@ object AzureApp extends LoaderApp[KafkaSourceConfig, KafkaSinkConfig](BuildInfo)
 
   override def badSink: SinkProvider = KafkaSink.resource(_, classTag[SinkAuthHandler])
 
-  override def isDestinationSetupError: DestinationSetupErrorCheck = TableFormatSetupError.check
+  override def isDestinationSetupError: DestinationSetupErrorCheck = {
+    // Authentication issue (wrong OAuth endpoint, wrong client id, wrong secret..)
+    case e: AbfsRestOperationException if e.getStatusCode == -1 =>
+      e.getErrorMessage()
```
benjben, line 45: Example with wrong endpoint:
```
Failed to create events table: Auth failure: HTTP Error 400; url='https://login.microsoftonline.com/0cf615b7-3054-4920-9991-58e992424671/oauth2/token' AADToken: HTTP connection to https://login.microsoftonline.com/0cf615b7-3054-4920-9991-58e992424671/oauth2/token failed for getting token from AzureAD.; requestId='cb17fcea-c326-4716-9c1f-f92c87765200'; contentType='application/json; charset=utf-8'; response '{"error":"invalid_request","error_description":"AADSTS90002: Tenant '0cf615b7-3054-4920-9991-58e992424671' not found. Check to make sure you have the correct tenant ID and are signing into the correct cloud. Check with your subscription administrator, this may happen if there are no active subscriptions for the tenant. Trace ID: cb17fcea-c326-4716-9c1f-f92c87765200 Correlation ID: 1e7caf23-be53-4aa8-991b-a458fd187357 Timestamp: 2024-10-18 13:52:56Z","error_codes":[90002],"timestamp":"2024-10-18 13:52:56Z","trace_id":"cb17fcea-c326-4716-9c1f-f92c87765200","correlation_id":"1e7caf23-be53-4aa8-991b-a458fd187357","error_uri":"https://login.microsoftonline.com/error?code=90002"}'
```
Wrong client id:
```
Failed to create events table: Auth failure: HTTP Error 400; url='https://login.microsoftonline.com/0cf615b7-3054-4920-9991-58e992424670/oauth2/token' AADToken: HTTP connection to https://login.microsoftonline.com/0cf615b7-3054-4920-9991-58e992424670/oauth2/token failed for getting token from AzureAD.; requestId='d594c7e6-60cb-43ab-b89b-687a87c31600'; contentType='application/json; charset=utf-8'; response '{"error":"unauthorized_client","error_description":"AADSTS700016: Application with identifier 'b1874e74-163f-4ca3-9551-16aa5c6460eb' was not found in the directory 'Snowplow'. This can happen if the application has not been installed by the administrator of the tenant or consented to by any user in the tenant. You may have sent your authentication request to the wrong tenant. Trace ID: d594c7e6-60cb-43ab-b89b-687a87c31600 Correlation ID: 899535f2-64a6-49ef-a256-0f3a75bf81a9 Timestamp: 2024-10-18 13:54:14Z","error_codes":[700016],"timestamp":"2024-10-18 13:54:14Z","trace_id":"d594c7e6-60cb-43ab-b89b-687a87c31600","correlation_id":"899535f2-64a6-49ef-a256-0f3a75bf81a9","error_uri":"https://login.microsoftonline.com/error?code=700016"}'
```
Wrong secret:
```
Failed to create events table: Auth failure: HTTP Error 401; url='https://login.microsoftonline.com/0cf615b7-3054-4920-9991-58e992424670/oauth2/token' AADToken: HTTP connection to https://login.microsoftonline.com/0cf615b7-3054-4920-9991-58e992424670/oauth2/token failed for getting token from AzureAD.; requestId='7baa91ac-4829-4d01-ad6a-bc5bc55b2800'; contentType='application/json; charset=utf-8'; response '{"error":"invalid_client","error_description":"AADSTS7000215: Invalid client secret provided. Ensure the secret being sent in the request is the client secret value, not the client secret ID, for a secret added to app 'b1874e74-163f-4ca3-9551-16aa5c6460ea'. Trace ID: 7baa91ac-4829-4d01-ad6a-bc5bc55b2800 Correlation ID: 05cc7746-12cc-4d48-9495-203803405b89 Timestamp: 2024-10-18 13:55:37Z","error_codes":[7000215],"timestamp":"2024-10-18 13:55:37Z","trace_id":"7baa91ac-4829-4d01-ad6a-bc5bc55b2800","correlation_id":"05cc7746-12cc-4d48-9495-203803405b89","error_uri":"https://login.microsoftonline.com/error?code=7000215"}'
```

istreeter:
Hi @benjben I am happy that customers now get _something_ if they do the setup flow wrong.  And I think your `checkHostname` is a clever way to solve the problem of not getting access to the original exception when the storage name is invalid.

My criticism is that the error messages are very messy in places where you call `e.getErrorMessage`.  Bearing in mind this error goes directly to the customer, currently it contains lots of "fluff" that doesn't add meaningful information to the customer:

- A request ID
- Content type of the error response
- A trace id
- A correlation ID
- An "error description" which contains some advice (which is not even good advice within the context BDP)

I know a clever customer could probably read this mess and extract the root problem (especially if they are familiar with azure).... but for a slick user-friendly BDP feature I think we can do better.

From the error messages you shared, it seems there is a `error_code` buried deep inside the string that tells us exactly what we need to know:

- If the error code is `90002` then customer provided the wrong tenant name
- If the error code is `700016` then customer provided the wrong client id or secret

I found [this Azure doc](https://learn.microsoft.com/en-us/entra/identity-platform/reference-error-codes) which lists all the error codes.  I also found [this resource](https://login.microsoftonline.com/error) where you can type in a error code to discover its meaning.

Is it possible to somehow extract that error code from the exception?  Either via `e.getErrorCode` or worst case write a parse to extract it out of `e.getErrorMessage`?

If we can achieve that, then the error messages sent to the customer can be really plain and simple, e.g. "Failed to create events table: Could not authenticate because Tenant ID is invalid."

benjben:
Hey @istreeter ,

> Is it possible to somehow extract that error code from the exception? Either via e.getErrorCode

Unfortunately for authentication errors the error code is `UNKNOWN`.

> If we can achieve that, then the error messages sent to the customer can be really plain and simple, e.g. "Failed to create events table: Could not authenticate because Tenant ID is invalid."

The number of error codes is enormous. What do you think of my approach :
- In case it is one of wrong tenant, wrong app ID, wrong secret, return a description of the error
- In case it is another error code, tell the customer to visit e.g. https://login.microsoftonline.com/error?code=90012 to get details about the error
- In case we can't extract the error code, return the full original message (should never happen)

?

istreeter:
I get very nervous about sending automated emails if we ourselves don't know the cause of the error.  For example.... is it possible to get a -1 status code for something which is our fault not the customer's fault?  I don't think we know that (unless you're certain you do know that???)

For that reason, I would lean towards:
- If it is wrong tenant, wrong app ID, wrong secret, return a description of the error
- If it is anything else... just don't send an error.  The exception will still get written to the logs.  The app will still crash and we will get other alerts because the loader is lagging.  Internally at Snowplow we will investigate the error, but we don't need to email the customer.

Over time, (if needed) we can add to the list of known error codes for which we choose to email the customer.

By the way.... interesting that you decided to extract the part of the error string that says `"error_description":"AADSTS90002:`.  In my earlier suggestion I was actually referring to a different part of the error string where it says `"error_codes":[90002]`.  I guess either is ok.  Do you have any opinions on which is better to extract with a regex?

benjben:
> For example.... is it possible to get a -1 status code for something which is our fault not the customer's fault? I don't think we know that

I'm not sure, but I agree that we should be cautious about that. For instance we could imagine something like the lake loader not having enough network permissions to contact the OAuth endpoint (just a stupid example).

> The exception will still get written to the logs. The app will still crash and we will get other alerts because the loader is lagging. Internally at Snowplow we will investigate the error, but we don't need to email the customer.

> Over time, (if needed) we can add to the list of known error codes for which we choose to email the customer.

I'm happy with this approach !

> I was actually referring to a different part of the error string where it says "error_codes":[90002]. I guess either is ok. Do you have any opinions on which is better to extract with a regex?

I don't have a strong opinion, I think that both work well. I didn't choose `"error_codes":[90002]` because it looks like a table that can have several values. But I could also just pick the first one. Would you prefer me to change? I'm easy!

modules/azure/src/main/scala/com.snowplowanalytics.snowplow.lakes/AuthenticationError.scala
```diff
@@ -0,0 +1,39 @@
...
+package com.snowplowanalytics.snowplow.lakes
+
+import org.apache.hadoop.fs.azurebfs.contracts.exceptions.AbfsRestOperationException
+
+import cats.implicits._
+
+object AuthenticationError {
+
+  private val knownErrors = Map(
+    90002 -> "Tenant is invalid. Make sure you have the correct tenant ID and are signing into the correct cloud. This may happen if there are no active subscriptions for the tenant",
```
istreeter, line 20: I think this part of the message...

> Make sure you have the correct tenant ID and are signing into the correct cloud. This may happen if there are no active subscriptions for the tenant

....is not phrased to be a helpful message for BDP customers.  I know you copied this from the original error message.  But I think it's the loader's job to just say "tenant is invalid", and then the console's job to wrap it by saying "here is the message from the loader, now go check the original setup instructions because you did it wrong".

tldr: I'd cut it to the first sentence only.

istreeter (commented):
Yay!!  I like the new changes.  I like the object with the custom `unapply`.

My last couple of suggestions are just minor tidy-up.

> I don't have a strong opinion, I think that both work well. I didn't choose "error_codes":[90002] because it looks like a table that can have several values. But I could also just pick the first one. Would you prefer me to change? I'm easy!

OK completely makes sense, let's stick with matching on the AADSTS string.  Do you think we should make the regex more restrictive by, say, adding word boundary matches?

```
val regex = """.*\bAADSTS(\d+)\b.*""".r
```

Anything else to make it more restrictive, without incurring potential future bugs!!

modules/azure/src/main/scala/com.snowplowanalytics.snowplow.lakes/AuthenticationError.scala
benjben: I kept `Tenant is invalid or does not have an active subscription`

benjben:
> Do you think we should make the regex more restrictive by, say, adding word boundary matches?

Good idea ! I added the word boundaries. I hesitated to add `error_description":"`before as well, but I'm afraid that the format might change one day. Do you think that I should add it ?

PR #91, opened 2024-10-25
PDP-1323 Handle initialization errors for GCP

benjben:


modules/gcp/src/main/scala/com.snowplowanalytics.snowplow.lakes/GcpApp.scala
```diff
@@ -19,5 +21,15 @@ object GcpApp extends LoaderApp[PubsubSourceConfig, PubsubSinkConfig](BuildInfo)
...
   override def badSink: SinkProvider = PubsubSink.resource(_)
 
-  override def isDestinationSetupError: DestinationSetupErrorCheck = TableFormatSetupError.check
+  override def isDestinationSetupError: DestinationSetupErrorCheck = {
+    // Destination bucket doesn't exist
+    case e: GoogleJsonResponseException if e.getDetails.getCode == 404 =>
+      "The specified bucket does not exist"
+    // Permissions missing for Cloud Storage
+    case e: GoogleJsonResponseException if e.getDetails.getCode == 403 =>
+      e.getDetails.getMessage
```
benjben, line 30: I hesitated between `getMessage` and `"Missing permissions for Google Cloud Storage"`.
When playing with my setup, I got an error with this content for `getMessage`:
```
ben-test-lake-loader@myproject.iam.gserviceaccount.com does not have storage.buckets.get access to the Google Cloud Storage bucket. Permission 'storage.buckets.get' denied       on resource (or it may not exist).
```
Thanks to this I could figure out the issue. With just the hard coded string I couldn't have.
That makes me in favor of `getMessage`.

modules/gcp/src/main/scala/com.snowplowanalytics.snowplow.lakes/GcpApp.scala
```diff
@@ -19,5 +21,15 @@ object GcpApp extends LoaderApp[PubsubSourceConfig, PubsubSinkConfig](BuildInfo)
 
   override def badSink: SinkProvider = PubsubSink.resource(_)
 
-  override def isDestinationSetupError: DestinationSetupErrorCheck = TableFormatSetupError.check
+  override def isDestinationSetupError: DestinationSetupErrorCheck = {
+    // Destination bucket doesn't exist
+    case e: GoogleJsonResponseException if e.getDetails.getCode == 404 =>
```
istreeter, line 26: I'm a big fan of cats triple equals in places like this

```scala
  case e: GoogleJsonResponseException if e.getDetails.getCode === 404 =>
```

istreeter (approved):
Looks good.

modules/gcp/src/main/scala/com.snowplowanalytics.snowplow.lakes/GcpApp.scala
benjben: I updated !

PR #94, opened 2024-10-31
Disable asynchronous deletes for Hudi

istreeter:
In #82 we added a feature to delete files asynchronously. This has worked great for Delta, which is tolerant to deletes that happen eventually not immediately. But it is not working great for Hudi, which does delete-and-replace of the `hoodie.properties` file.

This commit disables the feature again for Hudi only.


oguzhanunlu (approved):
👍 thanks!

PR #98, opened 2024-12-24
Disable AWS SDK deprecation warning

istreeter:
Newer minor versions of the v1 SDK print a warning to the logs:

```
The AWS SDK for Java 1.x entered maintenance mode starting July 31, 2024 and will reach end of support on December 31, 2025. For more information, see https://aws.amazon.com/blogs/developer/the-aws-sdk-for-java-1-x-is-in-maintenance-mode-effective-july-31-2024/
```

...plus a stack trace.

The Lake Loader uses the v1 SDK via a transitive depedency of Hadoop. The next release of Hadoop is moving to the v2 SDK, and the Lake Loader will follow Hadoop latest releases. Therefore it is OK to just disable the warning message for now.


project/BuildSettings.scala
```diff
@@ -67,7 +67,10 @@ object BuildSettings {
 
   lazy val awsSettings = appSettings ++ Seq(
     name := "lake-loader-aws",
-    buildInfoKeys += BuildInfoKey("cloud" -> "AWS")
+    buildInfoKeys += BuildInfoKey("cloud" -> "AWS"),
+
+    // TODO: Remove this after Hadoop 3.5.0 is released with full support for V2 SDK
```
benjben, line 72: We might want to put a comment next to Hadoop version as well

benjben (approved):

project/BuildSettings.scala
istreeter: I absolutely promise I will remember :)  This Hadoop release (+ removing aws sdk 1) is a big thing for Lake Loader, and I am following it closely.

PR #102, opened 2024-12-31
Repartition by event name before writing to the lake

istreeter:
Previously, our Iceberg writer was using the [hash write distribution mode][1] because that is the default for Iceberg. In this mode, Spark repartitions by the dataframe immediately before writing to the lake.

After this commit, we explicitly repartition the dataframe as part of the existing spark task for preparing the final dataframe. This means we can change the Iceberg write distribution mode to `none`.

We partition by the combination `event_name + event_id`: the former because it matches the lake partitioning, and the latter because it adds salt ensures equally sized partitions.

Overall this seems to improve the time taken to write a window of events to Iceberg. This fixes a problem we found, in which the write phase could get too slow when under high load (Iceberg only): specifically, a write was taking longer than the loader's "window" and this caused periods of low cpu usage, where the loader's processing phase was waiting for the write phase to catch up.

This commit also removes the config option `writerParallelismFraction`.  Before this commit, there were disadvantages to making the writer parallelism too high, because it would lead to smaller file sizes. But after this commit, now that we partition by event_name, we might as well make the writer parallelism as high as reasonably possible, which also speeds up the write phase of the loader.

Note: this improvement will not help Snowplow users who have changed the parition key to something different to our default. We might want to make a follow-up change, in which it auto-discovers the lake's partition key. For example, some users might want to partition by `app_id` instead of `event_name`.

[1]: https://iceberg.apache.org/docs/1.7.1/spark-writes/#writing-distribution-modes

### DIFF
--- a/modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriter.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriter.scala
@@ -81,10 +81,9 @@ object LakeWriter {
     }
     for {
       session <- SparkUtils.session[F](config, w, target.location)
-      writerParallelism = chooseWriterParallelism(config)
-      mutex1 <- Resource.eval(Mutex[F])
-      mutex2 <- Resource.eval(Mutex[F])
-    } yield impl(session, w, writerParallelism, mutex1, mutex2)
+      writerParallelism = chooseWriterParallelism()
+      mutex <- Resource.eval(Mutex[F])
+    } yield impl(session, w, writerParallelism, mutex)
   }
 
   def withHandledErrors[F[_]: Async](
@@ -129,22 +128,14 @@ object LakeWriter {
   /**
    * Implementation of the LakeWriter
    *
-   * The mutexes are needed because we allow overlapping windows. They prevent two different windows
+   * The mutex is needed because we allow overlapping windows. They prevent two different windows
    * from trying to run the same expensive operation at the same time.
-   *
-   * @param mutextForWriting
-   *   Makes sure there is only ever one spark job trying to write events to the lake. This is a
-   *   IO-intensive task.
-   * @param mutexForUnioning
-   *   Makes sure there is only ever one spark job trying to union smaller DataFrames into a larger
-   *   DataFrame, immediately before writing to the lake. This is a cpu-intensive task.
    */
   private def impl[F[_]: Sync](
     spark: SparkSession,
     w: Writer,
     writerParallelism: Int,
-    mutexForWriting: Mutex[F],
-    mutexForUnioning: Mutex[F]
+    mutex: Mutex[F]
   ): LakeWriter[F] = new LakeWriter[F] {
     def createTable: F[Unit] =
       w.prepareTable(spark)
@@ -164,10 +155,8 @@ object LakeWriter {
 
     def commit(viewName: String): F[Unit] =
       for {
-        df <- mutexForUnioning.lock.surround {
-                SparkUtils.prepareFinalDataFrame(spark, viewName, writerParallelism)
-              }
-        _ <- mutexForWriting.lock
+        df <- SparkUtils.prepareFinalDataFrame(spark, viewName, writerParallelism)
+        _ <- mutex.lock
                .surround {
                  w.write(df)
                }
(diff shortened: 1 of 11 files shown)


istreeter:
Shows the spark UI, for a load test with 4 cpu available:

spark-ui-20250109-aqe-disabled

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriter.scala
```diff
@@ -164,25 +155,21 @@ object LakeWriter {
 
     def commit(viewName: String): F[Unit] =
       for {
-        df <- mutexForUnioning.lock.surround {
-                SparkUtils.prepareFinalDataFrame(spark, viewName, writerParallelism)
-              }
-        _ <- mutexForWriting.lock
+        df <- SparkUtils.prepareFinalDataFrame(spark, viewName, writerParallelism)
```
spenes, line 158: Is it okay if `prepareFinalDataFrame` is run by two different windows from now on ?

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkUtils.scala
```diff
@@ -70,7 +70,11 @@ private[processing] object SparkUtils {
     for {
       _ <- Logger[F].debug(s"Initializing local DataFrame with name $viewName")
       _ <- Sync[F].blocking {
-             spark.emptyDataFrame.createTempView(viewName)
+             try {
+               spark.sparkContext.setLocalProperty("spark.scheduler.pool", "pool1")
```
spenes, line 74: If I understand correctly, `pool1` has higher priority than default pool. Are we using it for the light-weight jobs ?

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkUtils.scala
```diff
@@ -97,17 +105,21 @@ private[processing] object SparkUtils {
     viewName: String,
     writerParallelism: Int
   ): F[DataFrame] =
-    Sync[F].blocking {
+    Sync[F].delay {
       spark
         .table(viewName)
+        .repartitionByRange(writerParallelism, col("event_name"), col("event_id"))
```
spenes, line 111: Why `event_id` is needed exactly ?

spenes (approved):
Glad to hear it gives better results! Left a few questions to understand it better.

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriter.scala
istreeter: Yes.  After these changes, `prepareFinalDataFrame` does not actually do anything.  In Spark, DataFrames are lazy so they are not evaluated until needed.  Before these changes, we force-evaluated the final DataFrame by calling `.localCheckpoint()`.

For the same reason, you might have noticed that `prepareFinalDataFrame` no longer uses `Sync[F].blocking{ }`.

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkUtils.scala
istreeter: Yes exactly.  The lightweight jobs only.  I want to make sure they don't get stuck in a queue weighting for the slower jobs.

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkUtils.scala
istreeter: There is a danger that repartitioning leads to unequal sized partitions.  For example... imagine we are running on a pod with 8cpu, but this pipeline receives 70% page pings, and 30% of everything else.  One of the output partitions will contain 70% of all the incoming events.  And this is not the best way to make utilize the 8 cpu.  And it becomes more difficult to sink all the events within the timespan of one window.

By adding `event_id`, all of the output partitions have exactly equal size.

istreeter:
Another screenshot with 8 cpu, loading Delta format (the previous screenshot was Iceberg format)

Screenshot_20250110_141022

These profiles are good, because the long "save" jobs are nicely shorter than the 5 minute window duration.

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/SparkUtils.scala
spenes: I see. Also, is usage of `repartitionByRange` with `event_name` and `event_id` field correct ? What I understand from [this doc](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.repartitionByRange.html), partition column needs to be numeric to be able partition by range.

PR #103, opened 2024-12-31
DynamoDB table webhook messages

istreeter:
In #89 we added support for Delta's "S3 multi-cluster" mode, which uses a DynamoDB table to allow safe concurrent access by multiple loaders.

This PR extends the webhook feature, so the webhook receives a friendly error message if the DynamoDB table is mis-configured


benjben (approved):
We might want to add Snowman tests for the 2 new scenarios (we already have tests for all the other ones)

istreeter:
Hopefully, this is a temporary feature.  The next release of Hadoop is likely to support locking directly via S3: https://issues.apache.org/jira/browse/HADOOP-19256.  It is unclear yet how this will impact Delta, but I think it will mean we no longer need the DynamoDB table for locking.

For this reason, I have deliberately not opened any ticket to add snowman tests.  It doesn't feel like effort in the right place, when the future of this feature is uncertain.  I am closely watching the hadoop tickets, so I will re-assess that opinion in a couple of months.

PR #109, opened 2025-01-13
Revert to unsorted Dataframe for Delta and Hudi

istreeter:
In #102 I changed how we partition and sort the DataFrame before writing to the Lake.

The new partitioning is working well for all output formats. But the extra sort step is not strictly unnecessary for Delta and Hudi. This commit removes the sort step for Delta/Hudi only, to slightly improve cpu utilization.


istreeter:
Sharing a graph which demonstrates this has a measurable impact.  The metric shown is `processing_latency_millis`.  The left-hand-side (yellow) is after this current change.  The right-hand-side (green) is before this current change.  There was a controlled constant stream of incoming events, at ~156K events per second.  Loader was running with 4 available CPU.

The metric shows that the current PR improves the processing latency by a small but consistent amount.

processing-latency-millis

There was also a slight measurable improvement in cpu utilization.

spenes (approved):
👍

PR #111, opened 2025-02-12
Extract missing S3 IAM action for AWS app

oguzhanunlu:
This PR aims to improve initialization error message in case lake loader's role doesn't have a required S3 action in the iam policy. In case of not deducing any action, loader still fallback to the previous message.

An example S3Exception message
```
User: arn:aws:sts::...loader is not authorized to perform: s3:PutObject on resource: "arn:aws:s3:::lake-loader/events/metadata/uuid-....metadata.json" because no identity-based policy allows the s3:PutObject action (Service: S3, Status Code: 403, Request ID: req-id, Extended Request ID:ext-req-id)
```

### DIFF
--- a/modules/aws/src/main/scala/com.snowplowanalytics.snowplow.lakes/AwsApp.scala
+++ b/modules/aws/src/main/scala/com.snowplowanalytics.snowplow.lakes/AwsApp.scala
@@ -48,7 +48,7 @@ object AwsApp extends LoaderApp[KinesisSourceConfig, KinesisSinkConfig](BuildInf
       "S3 bucket does not exist or we do not have permissions to see it exists"
     case e: S3Exception if e.statusCode() === 403 =>
       // No permission to read from S3 bucket or to write to S3 bucket
-      extractS3Action(e)
+      extractMissingPermission(e)
     case e: S3Exception if e.statusCode() === 301 =>
       // Misconfigured AWS region
       "S3 bucket is not in the expected region"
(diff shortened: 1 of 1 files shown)


modules/aws/src/main/scala/com.snowplowanalytics.snowplow.lakes/AwsApp.scala
```diff
@@ -103,4 +105,15 @@ object AwsApp extends LoaderApp[KinesisSourceConfig, KinesisSinkConfig](BuildInf
...
+  /**
+   * Extracts s3:ActionName from exception message if possible
+   *
+   * Example exception message is as below: User: arn:aws:sts::...loader is not authorized to
+   * perform: s3:PutObject on resource: "arn:aws:s3:::bucket-name/....json" because no
+   * identity-based policy allows the s3:PutObject action (Service: S3, Status Code: 403, Request
+   * ID: req-id, Extended Request ID: ext-req-id)
+   */
+  private def extractS3Action(s3Exception: S3Exception): Option[String] =
+    s3Exception.getMessage.split(" ").filter(_.startsWith("s3:")).distinct.headOption
```
benjben, line 118: Matching only on `s3:` doesn't sound very strict. Also, it doesn't feel good to have `.distinct.headOption` and have a `fold` on the result. Here is an alternative with a regex:
```suggestion
  private def extractMissingPermission(e)(s3Exception: S3Exception): String = {
    val pattern = """.*is not authorized to perform: s3:(\w+).*""".r
    s3Exception.getMessage match {
      case pattern(action) =>
        s"Missing s3:$action permission to setup the destination table in S3 bucket"
      case _ =>
        "Missing permission to setup the destination table in S3 bucket"
  }
```
(not tested)

benjben (changes requested):

PR #117, opened 2025-04-08
Fix unlikely race condition in LakeWriter

istreeter:
Before #102, we had two mutexes in LakeWriter. In #102 I removed one of them (amongst other changes).  Now I realize that introduced a potential race condition, because of how we save, modify, and re-save a dataframe by name, and because we have parallelism over the that part of the code.

This commit re-introduces the mutex again, to avoid the potential race condition.

I never observed the race condition, and I think it was very unlikely. Introducing a mutex here should not impact performance.

### DIFF
--- a/modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriter.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriter.scala
@@ -82,8 +82,9 @@ object LakeWriter {
     for {
       session <- SparkUtils.session[F](config, w, target.location)
       writerParallelism = chooseWriterParallelism()
-      mutex <- Resource.eval(Mutex[F])
-    } yield impl(session, w, writerParallelism, mutex)
+      mutex1 <- Resource.eval(Mutex[F])
+      mutex2 <- Resource.eval(Mutex[F])
+    } yield impl(session, w, writerParallelism, mutex1, mutex2)
   }
 
   def withHandledErrors[F[_]: Async](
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriter.scala
```diff
@@ -128,14 +129,23 @@ object LakeWriter {
   /**
    * Implementation of the LakeWriter
    *
-   * The mutex is needed because we allow overlapping windows. They prevent two different windows
+   * The mutexes are needed because we allow overlapping windows. They prevent two different windows
    * from trying to run the same expensive operation at the same time.
+   *
+   * @param mutexForRemoteWriting
+   *   This mutex is needed because we allow overlapping windows. It prevents two different windows
+   *   from trying to run the same expensive operation at the same time
```
benjben, line 137: This documentation is the exact same as the one above, would it be possible to be a bit more specific, like for `mutexForLocalAppending`?

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriter.scala
```diff
@@ -128,14 +129,23 @@ object LakeWriter {
   /**
    * Implementation of the LakeWriter
    *
-   * The mutex is needed because we allow overlapping windows. They prevent two different windows
+   * The mutexes are needed because we allow overlapping windows. They prevent two different windows
    * from trying to run the same expensive operation at the same time.
```
benjben, line 133: If I understand correctly, the risk is that the second window wants to start doing the same operation while the first one is not over yet, right?

Is it "only" because the operations are expensive that we want only one at a time or also for other reasons?

benjben (approved):

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriter.scala
istreeter: My mistake, I meant to delete the description at the top.  This description here is already specific to this mutex only.

modules/core/src/main/scala/com.snowplowanalytics.snowplow.lakes/processing/LakeWriter.scala
istreeter: For `mutexForRemoteWriting`, it is only because the operation is expensive.  For `mutexForLocalAppending` it is nothing to do with expensive operations, it is only to prevent a race condition.

This is my mistake, I forgot to delete two lines, which has made everything more confusing.

benjben (approved):

